PR #50, opened 2014-01-30
Granular permissions

### LINKED ISSUE #49: Granular permissions
Is there an easy way to split permissions for execution and manipulation of queries? I'd like to give my staff users the right to execute saved queries but only admins should be able to change and save queries.

Thanks.

lociii:
Fixes #49

Make it easy to define who can access the sql explorer.
EXPLORER_PERMISSION_VIEW and EXPLORER_PERMISSION_CHANGE can be changed using django config settings and should provide a callback that accepts a user as the only argument and should return True or False.
The default is still that the user is required to be staff.

### DIFF
--- a/explorer/app_settings.py
+++ b/explorer/app_settings.py
@@ -12,14 +12,10 @@ EXPLORER_CONNECTION_NAME = getattr(settings, 'EXPLORER_CONNECTION_NAME', None)
 
 
 def sql_explorer_view(user):
-    if user.is_staff:
-        return True
-    return False
+    return user.is_staff
 EXPLORER_PERMISSION_VIEW = getattr(settings, 'EXPLORER_PERMISSION_VIEW', sql_explorer_view)
 
 
 def sql_explorer_change(user):
-    if user.is_staff:
-        return True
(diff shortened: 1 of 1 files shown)


amiraliakbari:
> I did this by writing my own decorator based on staff_member_required that rewrote the page as an admin login page, rather than redirecting a login page as user_passes_test does.

Just to note that this may cause problems in projects that want to have a centralized login page. For example we have a login page that requires a captcha for login, but the login in explorer app create a way to bypass it. I think including login in reusable apps is not a good idea.

I use this code in project's `urls.py` to disable it:

    from django.views.generic import RedirectView
    from explorer import utils as explorer_utils

    explorer_utils.safe_login_prompt = RedirectView.as_view(url='/login')

As this behavior is not explicitly inferred from documentation, I think it will be helpful to provide an option for this.

chrisclark:
Hey @amiraliakbari I'm not sure I understand -- the safe_login_prompt helper sends the user to the default admin login (using admin/login.html). Do you mean an option for redirecting to the default Django *account* login?

PR #95, opened 2014-12-11
Added support for Python 2.6 with a custom counter class from the Intern...

rodvdka:
...et with no license restrictions and a small change to views.py

Counter.py is from http://code.activestate.com/recipes/576611-counter-class/ with an MIT License.

This should make it run without any issues on Python 2.6.

### DIFF
--- a/explorer/counter.py
+++ b/explorer/counter.py
@@ -1,0 +1,184 @@
+from operator import itemgetter
+from heapq import nlargest
+from itertools import repeat, ifilter
+
+class Counter(dict):
+    '''Dict subclass for counting hashable objects.  Sometimes called a bag
+    or multiset.  Elements are stored as dictionary keys and their counts
+    are stored as dictionary values.
+
+    >>> Counter('zyzygy')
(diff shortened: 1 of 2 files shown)


sindrig:
I did a similar thing before seeing this pull request.

chrisclark:
Sounds like there is appetite for this. And @rodvdka I'm sorry for ignoring this for so long.

Could you strip down the counter class so it looks more like this?

https://github.com/epantry/django-sql-explorer/compare/compat-26

I hate having so much dead and untested code hanging around. I can just merge that branch as well, but I'd love to give you credit for the contribution.

rodvdka:
Thanks, I'll do it tomorrow at work.

chrisclark:
Let me know if you want me to do it as well. I can still add you to the contributors file :)

rodvdka:
Please confirm this commit now. Thanks @chrisclark

chrisclark:
0.7 has just been released which includes this change. Thanks for the help!

PR #106, opened 2015-03-25
Python3 support

leebrooks0:
Hi

I got the tests to work on both 2.7 and 3.4

I will squash commits and remove the 2 TODO comments and submit a new PR.

Just checking that you are happy with things so far?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: python
 python:
+  - "2.6"
   - "2.7"
   - "3.4"
 install: pip install -r requirements.txt
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,5 +3,9 @@ python:
   - "2.6"
   - "2.7"
   - "3.4"
-install: pip install -r requirements.txt
+install:
+  - pip install -r requirements.txt
+  # Python 2.6 is only supported until Django 1.6
+  - if [[ $TRAVIS_PYTHON_VERSION == '2.6' ]]; then pip install django=1.6.7; fi
+
 script:  python manage.py test --settings=explorer.tests.settings
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,6 @@
 language: python
 python:
   - "2.7"
-  - "3.2"
-  - "3.3"
   - "3.4"
 install: pip install -r requirements.txt
 script:  python manage.py test --settings=explorer.tests.settings
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,8 @@
 language: python
 python:
   - "2.7"
+  - "3.2"
+  - "3.3"
+  - "3.4"
 install: pip install -r requirements.txt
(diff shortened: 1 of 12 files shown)


.travis.yml
```diff
@@ -1,5 +1,6 @@
 language: python
 python:
   - "2.7"
+  - "3.4"
```
chrisclark: We actually target Python 2.6 compatibility as well -- see this commit:
https://github.com/epantry/django-sql-explorer/commit/8b445b63fcd2d3dd1cb6bd36df119310c4112a25

Probably should have added that into Travis as well. Can you make sure 2.6 still runs and add it to travis?

explorer/actions.py
```diff
@@ -37,7 +37,10 @@ def _package(queries):
 def _build_zip(queries):
     temp = tempfile.TemporaryFile()
     zip_file = ZipFile(temp, 'w')
-    map(lambda r: zip_file.writestr('%s.csv' % r.title, csv_report(r) or "Error!"), queries)
+    # TODO: Work out why this doesn't work but the latter does
+    # map(lambda r: zip_file.writestr('%s.csv' % r.title, csv_report(r) or "Error!"), queries)
+    for r in queries:
+        zip_file.writestr('%s.csv' % r.title, csv_report(r) or "Error!")
```
chrisclark: Yeah, this was too clever anyway. I was in my 'purely functional python' phase.

This fails in python3 because map returns an iterator instead of a list (http://www.diveintopython3.net/porting-code-to-python-3-with-2to3.html#map), so evaluation never happens. It's bad practice to use map() for side-effects anyway, so this is really a solid change. Feel free to remove the comment.

explorer/utils.py
```diff
@@ -85,10 +85,13 @@ def extract_params(text):
...
 def write_csv(headers, data):
-    csv_data = cStringIO.StringIO()
+    csv_data = cStringIO()
     writer = csv.writer(csv_data)
     writer.writerow(headers)
-    map(lambda row: writer.writerow(row), data)
+    # TODO: Resolve why this line doesn't work on Py 3.4
+    # map(lambda row: writer.writerow(row), data)
+    for row in data:
+        writer.writerow(row)
```
chrisclark: Same here.

requirements.txt
```diff
@@ -1,3 +1,5 @@
-django>=1.6.7
+Django==1.7.7
```
chrisclark: This is a big move. Does this break django 1.6 compat?

chrisclark:
Looks pretty good -- thanks! Can you add Python 2.6 to the travis file to confirm it still works?

The only thing I take issue with is the Django 1.7 requirement. Many folks haven't made that jump yet and I'd like to avoid that dependency if possible.

leebrooks0:
Thanks, I will wrap up tomorrow morning and add 2.6 to Travis. Django 1.7.
was actually a pip freeze > requirements.txt and it "slipped" in.

PR #120, opened 2015-06-14
Fix for showing active page in header for logs

dsanders11:
Minor visual fix


chrisclark:
If you have a minute, would you mind rebasing this from master so I don't have to include a merge commit? You can also add yourself to the AUTHORS file :)

dsanders11:
@chrisclark, I believe this got merged in with #121 so you I'm just going to close this.

PR #121, opened 2015-06-14
Use url template tag instead of relative urls

dsanders11:
This originally started as a fix for the link in the navbar brand (it is currently '../' so on the main query list it takes you out of the explorer), but I went ahead and went through updating any URLs I found in the templates to use the `url` template tag. It makes it easier for end users to customize the URLs via the urlconf without having all of the templates break.

There are still hardcoded URLs in `explorer.js` but those are trickier to change.


chrisclark:
This is great, thanks!

PR #140, opened 2015-10-06
Test the different Django versions

grantmcconnaughey:
To ensure compatibility, test the 3 versions of Django that the README says is supported.


grantmcconnaughey:
With this, you can add Django versions (like 1.9a1) and Python versions (3.2, 3.3, 3.5) and then brag about supporting them on your README.rst!

chrisclark:
This is totally awesome -- I know nothing about Travis so super helpful. Thanks!

PR #146, opened 2015-11-24
Avoid converting to UTF-8 in CSV generation

### LINKED ISSUE #145: Text columns now export to CSV as bytestring representation
Downloading a query result as CSV when using Python 3/MySQL now results in text columns being bytestrings, i.e. `b'Example'` rather than just `Example`.

1. Was this change intentional?
1. Is this configurable or is there a workaround? I thought of using transforms, but I don't think I'll be able to achieve what I'm looking to do.

daneah:
Fixes #145

### DIFF
--- a/explorer/models.py
+++ b/explorer/models.py
@@ -150,11 +150,6 @@ class QueryResult(object):
             return [ix for ix, _ in enumerate(self._description) if not isinstance(d[ix], six.string_types) and six.text_type(d[ix]).isnumeric()]
         return []
 
-    def _get_unicodes(self):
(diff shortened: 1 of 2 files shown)


daneah:
@chrisclark any movement on this?

chrisclark:
I'll merge this in and try it out for a bit -- if this works, that would be terrific because it removes a bunch of silliness around unicode handling. Thanks!

daneah:
@chrisclark indeed, I try to avoid unicode as much as possible :smile:

PR #153, opened 2015-12-07
Removed deprecated future url template tag for Django 1.9

johanndt:
When running under Django 1.9, the following exception was being thrown: `TemplateSyntaxError: 'url' is not a valid tag or filter in tag library 'future'`

The new url tag syntax has been with us since Django 1.5, and django-sql-explorer claims to only support Django 1.6+, so the solution was to simply remove the offending import in the templates.


chrisclark:
Thanks very much!

PR #155, opened 2015-12-11
Remove instances of {% load url from future %}

### LINKED ISSUE #156: {% load url from future %} no longer needed
`{% load url from future %}` was suggested for Django versions 1.3-1.5 as a way to extend the variety of types that `{% url %}` would accept. That has been deprecated or removed in all versions of Django that `django-sql-explorer` now supports. In cases where it's been removed, it breaks the template rendering.

From [Django 1.5 release notes](https://docs.djangoproject.com/en/1.9/releases/1.5/):
The upshot of this is that if you are not using `{% load url from future %}` in your templates, you’ll need to change tags like `{% url myview %}` to `{% url "myview" %}`. If you _were_ using `{% load url from future %}` you can simply remove that line under Django 1.5

From [Django 1.7 release notes](https://docs.djangoproject.com/en/1.9/releases/1.7/):
Django 1.3 introduced `{% load ssi from future %}` and `{% load url from future %}` syntax for forward compatibility of the `ssi` and `url` template tags. This syntax is now deprecated and will be removed in Django 1.9. You can simply remove the `{% load ... from future %}` tags.

daneah:
Resolves #156

### DIFF
--- a/explorer/templates/explorer/base.html
+++ b/explorer/templates/explorer/base.html
@@ -1,4 +1,3 @@
-{% load url from future %}
 {% load staticfiles %}
 <!DOCTYPE html>
 <html>
--- a/explorer/templates/explorer/play.html
+++ b/explorer/templates/explorer/play.html
@@ -1,7 +1,5 @@
 {% extends "explorer/base.html" %}
 
-{% load url from future %}
-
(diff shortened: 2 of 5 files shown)


chrisclark:
Yep. Thanks!

PR #159, opened 2015-12-11
Fix query list titles

### LINKED ISSUE #158: Query titles inconsistent on query list page
While query titles show up in the "All Queries" panel, they are replaced by "Query object" in the "Most Recently Used" panel. This likely has to do with `Query`'s `__unicode__` vs using `__str__`, etc.

daneah:
Fixes #158


daneah:
Sorry, this branch is dirty as it includes unmerged commits that are not part of this issue.

PR #163, opened 2016-01-05
Add Django 1.9, Python 3.5 to travis.yml

grantmcconnaughey:


grantmcconnaughey:
Not sure what to do about this. :\ It's not going to work on Django 1.9 while that djcelery requirement is there. djcelery doesn't seem to be Django 1.9 compatible (in fact, I don't think that app is maintained anymore).

chrisclark:
Thanks @grantmcconnaughey. I will do some digging and try to figure out a good approach.

DJCelery is not maintained, but I still find it useful -- sounds like 1.9 will be the forcing function get off it though. I expect to tackle this upgrade in the next 45 days. Not sure what to do either until I have a chance to dig in :)

chrisclark:
I have a solution -- look at the 'deprecations' branch that is in progress. I will get it merged over the weekend and then we should be good to go for this combo. Though notably I am dropping support for Django 1.6 and Python 2.6.

chrisclark:
I'm closing this PR because I just merged the 'deprecations' branch which addresses this.

Issue #94, opened 2014-11-20
Implement optimistic locking on queries


chrisclark:

lawson89:
@marksweb given how old this issue is and the fact that optimistic locking is for updates/deletes (at least I thought). I vote we close it and revisit if more detail on a use case comes up

PR #164, opened 2016-01-06
Add unsaved changes message

### LINKED ISSUE #45: "You have unsaved changes" message

grantmcconnaughey:
Fixes #45 

A "You have unsaved changes" message pops up if the user has changed the SQL. It only pops up if they try to go to another page but not if they click the "Save and Run" button. Also, it doesn't pop up when you change the SQL on a playground and then go to another page. I figured the playground isn't usually for saving queries anyway, so why annoy the user?

Here's what it looks like:

screenshot 2016-01-05 22 28 06


chrisclark:
Terrific -- thanks for this Grant. Simple and effective!

Next up -- issue #94!

PR #166, opened 2016-01-06
Add missing migration for South

tzangms:
I've a Django still using 1.6, so I encountered a problem with schema change of django-sql-explorer.

So I generate a missing migration file for South.


grantmcconnaughey:
If this is for South then it needs to go in the `south_migrations` directory, not the `migrations` directory. That's why the Travis CI tests are failing.

tzangms:
@grantmcconnaughey I think I put it in the wrong directory, I will submit a new PR, thanks

PR #168, opened 2016-01-06
Django version downgraded

kunwarakash:
Django 1.9 gives 'url' is not a valid tag library.
Install Django 1.8.3

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-Django>=1.6.7
+Django==1.8.3
 factory-boy==2.6.0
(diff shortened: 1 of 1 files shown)


kunwarakash:
After downgrading django to 1.8.3 issue is resolved

chrisclark:
I'm not sure I understand this PR. I guess there might be an issue with the url library? But can you file that as an issue instead of this PR? I don't think I'll merge this as it just targets 1.8.3 specifically, which I don't want to do.

PR #169, opened 2016-01-08
Fix missing South migration

tzangms:
Fix missing South migration for Django before 1.7 which still use South.


chrisclark:
Thanks!

tzangms:
yeah!

PR #173, opened 2016-01-16
Added django permission support for table access restriction

sharmashobhit:
- Added the functionality to give the user access to particular tables based on django permissions.

Usage:
Set EXPLORER_TABLE_LEVEL_PERMISSION = True in settings.py

In order to give the users access to a particular table, just give them the access to explorer by adding them to a group(s):
EXPLORER_PERMISSION_VIEW = lambda u: u.is_superuser or u.groups.filter(name__in=["SQL_EXPLORER_USERS1", "SQL_EXPLORER_USERS2", "SQL_EXPLORER_USERS3"]).exists()

And give table specific permission to the entire group.

You can also keep one group for user access to explorer and then give table access to the user.

P.S. Right now I've given the access on "change_<model_name>" permission.

### DIFF
--- a/explorer/app_settings.py
+++ b/explorer/app_settings.py
@@ -27,3 +27,6 @@ S3_ACCESS_KEY = getattr(settings, "EXPLORER_S3_ACCESS_KEY", None)
 S3_SECRET_KEY = getattr(settings, "EXPLORER_S3_SECRET_KEY", None)
 S3_BUCKET = getattr(settings, "EXPLORER_S3_BUCKET", None)
 FROM_EMAIL = getattr(settings, 'EXPLORER_FROM_EMAIL', 'django-sql-explorer@example.com')
+
+# Table level permissions
+EXPLORER_TABLE_LEVEL_PERMISSION = getattr(settings, "EXPLORER_TABLE_LEVEL_PERMISSION", False)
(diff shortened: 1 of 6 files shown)


sharmashobhit:
Modified requirements.txt as models.get_apps() has been deprecated in >=Django 1.9.0

chrisclark:
Hey @sharmashobhit -- thanks so much for this. Love the idea.

I've looked through it and need to noodle on it to think about how it works before I merge. I don't anticipate any problems *except* that Explorer does now support Django 1.9 (and dropped support for 1.6). You should be able to see the new way to access get_apps() by looking through: https://github.com/epantry/django-sql-explorer/pull/177

One other question -- why did you go with "change_" as the permission? Since explorer is (I assume) mostly used as a read-only tool, wouldn't another permission be more suitable? I haven't thought deeply about this, so could be wrong.

chrisclark:
I've thought about this more and just wanted to update you --

The only thing (I think) currently tying Explorer to Django is the Schema Explorer functionality. Otherwise, it is a Django-powered, but DB-agnostic query tool. There's no particular reason it has to be interrogating a Django DB; if you set it up with a connection to a non-django database, everything works perfectly (except the schema stuff).

I would love to re-write the schema feature to be application agnostic, which would really increase the reach of the tool. This PR, while incredibly useful, moves Explorer in the other direction; being more tightly coupled to querying a Django application database.

Additionally, I don't love relying on sql parsing for security (the blacklist is there for convenience, but I don't even particularly like that). I really do believe that this sort of thing should be (and technically, *has* to be) enforced at the DB-level.

That said, this is a really useful feature, so I'm still noodling on how to get a variation into master, but I don't see exactly how to do it at the moment.

chrisclark:
I'm working on making the tool DB agnostic via #247 so am going to close this PR.

PR #179, opened 2016-02-02
Update setup.py

millin:
To fix installation problems


chrisclark:
Thanks! I can do a point release to address.

chrisclark:
Done. Just released 0.9.2 with this patch.

PR #184, opened 2016-03-02
Update list.js to 1.2.0

kalekseev:
With list.js 1.1.1 I experienced a problem sorting columns with numbers (on sort column was filled with some random numbers from original data) with 1.2.0 there is no such problem.


chrisclark:
Looks great! Appreciate the contribution.

PR #188, opened 2016-03-23
Add initial exporters

chrisclark:
Opening a PR for comments and review.

### DIFF
--- a/explorer/exporters.py
+++ b/explorer/exporters.py
@@ -1,4 +1,8 @@
 import json
+try:
+    import cStringIO as StringIO
+except ImportError:
+    import StringIO
 
 from django.http import HttpResponse
 from django.utils.module_loading import import_string
@@ -16,35 +20,36 @@ class BaseExporter(object):
 
     name = ''
     content_type = ''
+    file_extension = ''
 
     def __init__(self, query):
         self.query = query
 
-    def to_response(self):
+    def get_output(self):
         raise NotImplementedError
+
+    def get_filename(self):
+        return '{}{}'.format(get_filename_for_title(
+            self.query.title), self.file_extension)
 
 
 class CSVExporter(BaseExporter):
 
     name = 'CSV'
     content_type = 'text/csv'
+    file_extension = '.csv'
 
-    def to_response(self):
-        data = csv_report(self.query, app_settings.CSV_DELIMETER).getvalue()
-        response = HttpResponse(data, content_type=self.content_type)
-        response['Content-Disposition'] = 'attachment; filename="%s.csv"' % (
-            get_filename_for_title(self.query.title)
-        )
-        response['Content-Length'] = len(data)
-        return response
+    def get_output(self):
+        return csv_report(self.query, app_settings.CSV_DELIMETER).getvalue()
 
 
 class JSONExporter(BaseExporter):
 
     name = 'JSON'
     content_type = 'application/json'
+    file_extension = '.json'
 
-    def to_response(self):
+    def get_output(self):
         res = self.query.execute_query_only()
         data = []
         for row in res.data:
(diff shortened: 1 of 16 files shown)


explorer/templatetags/explorer_tags.py
```diff
@@ -0,0 +1,19 @@
...
+@register.inclusion_tag('explorer/export_buttons.html')
+def export_buttons(query=None):
+    exporters = {}
+    for key, value in app_settings.EXPLORER_DATA_EXPORTERS.items():
+        exporter_class = import_string(value)
+        exporters[key] = exporter_class.name
+    return {
+        'exporters': exporters,
+        'query': query,
+    }
```
chrisclark: This is cool. From a UI perspective I was imagining a single 'Download' button, with a drop-down box to select the format, but let's give this a try. If it's clunky, maybe we revisit.

explorer/exporters.py
```diff
@@ -0,0 +1,99 @@
...
+class BaseExporter(object):
+
+    name = ''
+    content_type = ''
+
+    def __init__(self, query):
+        self.query = query
+
+    def to_response(self):
+        raise NotImplementedError
```
chrisclark: What do you think of putting more logic into the base class? I think it would increase cohesion, and allow us to remove some orphaned functions from utils. Something like (totally untested):

```py
class BaseExporter(object):

    name = ''
    content_type = ''

    def __init__(self, query):
        self.query = query

    def get_data(self, **kwargs):
        raise NotImplementedError

    def get_filename(self):  # inlined utils.get_filename_for_title()
        valid_chars = '-_.() %s%s' % (string.ascii_letters, string.digits)
        filename = ''.join(c for c in self.query.title if c in valid_chars)
        filename = filename.replace(' ', '_')
        return filename

    def to_response(self, download=True, **kwargs):
        data = get_data(self.query, **kwargs)
        response = HttpResponse(data, content_type=self.content_type)
        response['Content-Length'] = len(data)
        if download:
            response['Content-Disposition'] = 'attachment; filename="%s.csv"' % self.get_filename()
        return response
```

Then CSVExporter just becomes (after inlining write_csv and csv_report from utils):

```py
class CSVExporter(BaseExporter):
    name = 'CSV'
    content_type = 'text/csv'

    def get_data(self, delim=settings.CSV_DELIMITER):
        try:
            res = self.query.execute_query_only()
        except DatabaseError as e:
            resp = cStringIO()
            return resp.write(str(e))  # consistent return type
        
        delim = '\t' if delim == 'tab' else delim
        csv_data = cStringIO()
        if PY3:
            writer = csv.writer(csv_data, delimiter=delim)
        else:
            writer = csv.writer(csv_data, delimiter=delim, encoding='utf-8')
        writer.writerow(res.headers)
        for row in res.data:
            writer.writerow([s for s in row])
        return csv_data.get_value()
```

This contains all of the gnarliness around specific formats to one method in each exporter, decoupling the HttpResonse bits from the format handling. You could imagine adding a `to_file()` method on the BaseExporter, for instance.

With all that said, someone smart once said "[don't make classes that end in -er](http://objology.blogspot.com/2011/09/one-of-best-bits-of-programming-advice.html)", so another option to consider is keeping the "response-building" methods on the views as you suggested in your comment, and instead adding `.to_<format>()` methods to the Query class itself. It doesn't have the same sex-appeal as the multiple-classes + exporter 'registry', but is in many ways a more SOLID approach.

chrisclark:
This is looking good! A bunch of comments as I was looking through it (I can make some of these changes myself -- not trying to stick you with all the work).

1. Need an answer for custom delimiters in CSV files
2. Code duplication in views.py -- a few functions that do essentially the same thing (construct an export response). Be good to factor this out.
3. Test coverage needed on some of the exporters and endpoints.
4. Extraneous playground download code needs to be removed (JS that wires up the buttons + old /csv endpoint)
5. I have mixed feelings about the templatetag implementation; on one hand, it's elegant and clearly a great way to do it. But on the other, it runs counter to the 'view_model' pattern that the project has been using to date. Explorer has outgrown that pattern, but I think that should probably be revisited separately, and I'm hesitant to introduce a new UI-building concept for this feature.

PR #189, opened 2016-03-29
Data exporters

chrisclark:
@grantmcconnaughey im not really sure if this is the way to do it, but I pulled your branch, did some some refactoring and pushed it up here.

Addressed a couple of things:

- Removes a bunch of old dead exporting code and replaced it with your new stuff
- Inlines a bunch of the CSV stuff instead of burying it utilities
- Refactored the base exporter class for consistent error handling
- Refactored the export functions in views.py to share more code
- Started getting tests to pass with the new API

Still plenty to do, but getting there quickly!

Known issues:

- Delimiters in CSV files
- actions.py -- do we even still want to support this? Not clear if there is much value in the admin interface.
- failing tests
- Testing out the export button template tag, when users have different view/change query permissions. Does it behave right?
- More tests on different export types


grantmcconnaughey:
Great! I'm liking your changes. Definitely makes the exporters/export views more generic.

A few thoughts:

* Delimiters - I'm not sure how to handle this, really. At least not on an export-by-export basis. I figured we could just use the global `CSV_DELIMITER` setting each time. So if users change it to `'|'` then it will export using the pipe character each time.
* actions.py - I'm pretty sure this can be removed, unless you want to add all of the data exporters to the admin list pages, which might be kind of cool.
* Export buttons - I'll look into this. I didn't consider permissions when I initially created it.
* Tests - Yep, hadn't made it that far yet.

chrisclark:
I think delimiters are pretty straightforward to handle -- if not the most elegant :)

I just made a commit that I believe will solve it (I laid the groundwork for it earlier by allowing **kwargs in the exporters' get_output methods).

Apparently there is not test coverage on the delimiter stuff...I will add later today.

chrisclark:
Alrighty -- quick update:

1. Delimiters are in and tested.
2. Tests are all passing (except for...)
3. Placeholder test for ExcelExporter is not working

I'm also running into an issue locally where the csrf token is not rendering ( ```{% csrf_token %}``` is doing nothing). Might be just me...I'll keep digging.

chrisclark:
Alright -- ready to rock! I'm going to test out in production at our company for a couple of days, and then prepare to release. 1.0??

grantmcconnaughey:
Hey, sorry for the delayed response. I got a new job about a month ago so I've been pretty busy lately. Looking forward to version 1.0!

chrisclark:
Awesome! Congrats on the new gig. Looking forward to testing this week.

On Mon, Apr 25, 2016 at 6:18 PM, Grant McConnaughey <
notifications@github.com> wrote:

> Hey, sorry for the delayed response. I got a new job about a month ago so
> I've been pretty busy lately. Looking forward to version 1.0!
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/groveco/django-sql-explorer/pull/189#issuecomment-214579033>
>



-- 

Chris Clark │ Grove Co. │ 617-571-7327

1770 Union St, San Francisco, CA

chrisclark:
Good to go! We've been using in production successfully for about a week. Definitely found a number of bugs, but they've been sorted out and this sucker is ready to merge.

grantmcconnaughey:
@chrisclark: Awesome, man. Can't wait to play around with it. Great work!

PR #192, opened 2016-04-23
Use RequestContext when rendering views to ensure CSRF token is set

dsanders11:
I just upgraded to the latest version and found that I got CSRF errors on most POST requests for the SQL explorer. It looks like the root cause is that `RequestContext` wasn't being used most of the time meaning that the context processor which sets the value for the CSRF token was not being set. I've updated the code and I think added `RequestContext` in all places where it needs to be used.

BTW, it's a bit concerning that CSRF protection was broken out of the box. I would hope most people are running it and this would have popped up sooner.

### DIFF
--- a/explorer/views.py
+++ b/explorer/views.py
@@ -3,6 +3,7 @@ import six
 
 from django.http.response import HttpResponseRedirect
 from django.shortcuts import render_to_response, get_object_or_404
+from django.template import RequestContext
 from django.views.generic.base import View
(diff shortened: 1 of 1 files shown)


chrisclark:
Yep -- definitely an issue. My fault when I attempted to fix a Django 1.10 deprecation warning. I'm fixing it in the data_exporters branch, which will be merged soon -- but happy to merge this in the mean time so master is working. Thanks!

PR #196, opened 2016-05-11
Trim unused code

dsanders11:
Presumably this code slipped through from some testing, I don't see any use for it.


chrisclark:
@dsanders11 that's actually a critical function used by the Explorer model metaclasses to enable query introspection. I'll be very surprised if tests pass after this change.

Just kidding. Thanks!

PR #199, opened 2016-05-12
Drop Counter file since only Python 2.7+ is supported

dsanders11:
Now that Python 2.6 is no longer supported the fallback Counter file isn't needed anymore. `collections.Counter` is available starting in Python 2.7.


chrisclark:
Great catch - thanks.

PR #201, opened 2016-05-19
Fixed an error with unicode queries on Python2

tuffnatty:
Without this patch, any query containing non-ascii characters crashes with UnicodeEncodeError.


chrisclark:
Thanks!

PR #202, opened 2016-05-19
Fixed forced lowercasing of queries

tuffnatty:
Without this patch, explorer lowercases queries although they may be case-sensitive (in my case, there were mixed-case column names within double quotes in PostgreSQL; but probably it also affected queries like `WHERE some_column = 'someMixedCaseValue'`)

### DIFF
--- a/explorer/utils.py
+++ b/explorer/utils.py
@@ -75,8 +75,8 @@ def param(name):
 def swap_params(sql, params):
     p = params.items() if params else {}
     for k, v in p:
-        regex = re.compile("\$\$%s(?:\:([^\$]+))?\$\$" % str(k).lower())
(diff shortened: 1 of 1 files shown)


chrisclark:
Can you provide a test that covers this case? Thanks!

tuffnatty:
Sorry, too busy to write tests right now ;) Maybe later. But I assure you that implicit lowercasing of queries including quoted strings is wrong.

chrisclark:
No problem. I will merge and add the test.

chrisclark:
https://github.com/groveco/django-sql-explorer/commit/68c6c71ffb284dc70cd327d6332d40329c0d0137

All set. Thanks!

PR #203, opened 2016-05-20
Add missing migration

marcgibbons:


chrisclark:
...how has this been missing for so long? Very weird.

PR #205, opened 2016-06-01
allow non staff login if passed by permissions

vinayinvicible:
Any user should be able to login, if explorer permission functions return True and pass the user.


vinayinvicible:
@chrisclark can you take a look at this

chrisclark:
This looks great, and is a terrific idea. Merging! Thanks!

PR #218, opened 2016-07-28
fix #213 - query log table overlapping columns

### LINKED ISSUE #213: UI Bug: Logs View, Columns Run By and Duration have their values overlap
If you notice the **Run By** and **Display** columns, their values below run into each other. That's all here.

screen shot 2016-07-12 at 12 03 45 pm

jproffitt:
The main fix I did here was remove the `table-layout: fixed` css style. That was causing columns to overlap.

I also added the bootstrap `.table-responsive` div to make the table scroll horizontally when on small screens, rather than being truncated on the right.

Lastly, I added a `.container` div around the footer so that the body content would not extend past the browser window, causing an ugly gap on the right side of the window.

### DIFF
--- a/explorer/static/explorer/explorer.css
+++ b/explorer/static/explorer/explorer.css
@@ -10,10 +10,6 @@
     transition: height 0.01s ease;
 }
 
-.query-list {
-    table-layout: fixed;
-}
-
 .rows-input {
     text-align: center;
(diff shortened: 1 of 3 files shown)


chrisclark:
I'm running out of ways to say thanks. Muchas gracias!

PR #219, opened 2016-07-28
Fix #215 - schema view overlap

### LINKED ISSUE #215: UI Bug: Long field types extend beyond view
Relatively obvious once you see it, the side viewer doesn't allow the field types to wrap below when they extend beyond the view.

screen shot 2016-07-12 at 12 12 26 pm

jproffitt:
This fixes a few style issues with the schema sidebar. I added `.table-responsive` divs to the schema tables to scroll horizontally if they are too wide. Also fixed some spacing issues with the iframe itself.

### DIFF
--- a/explorer/static/explorer/explorer.css
+++ b/explorer/static/explorer/explorer.css
@@ -104,3 +104,7 @@ span.sort {
 .pvtAxisContainer, .pvtVals {
     border-color: #ddd;
 }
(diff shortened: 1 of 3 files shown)


chrisclark:
Makes sense. I've been keeping up with Bootstrap versions, but not actually with the best practices :)

PR #220, opened 2016-07-28
Fix #211 - save schema open status in cookie, other schema opening fixes

### LINKED ISSUE #211: UI Enhancement: Allow Users to Pin Open Schema
**Situation**: when a user searches a query in the playground or other views, the Schema viewer is automatically hidden, because the page reloads. This can lead to the user re-opening the Schema viewer multiple times as they try out a query.

**Suggested Solution**: Let this be set on a per session basis at least with javascript with a checkbox right above the schema viewer that says something like, "Keep Open".

jproffitt:
This PR adds support to "save" the state of the schema side bar in a cookie, so that the next time you load the page, it will remain open (or closed).

I also did some other cleanup related to opening the schema sidebar. There was some incorrect bootstrap grid structure.

### DIFF
--- a/explorer/static/explorer/explorer.js
+++ b/explorer/static/explorer/explorer.js
@@ -32,6 +32,10 @@ function ExplorerEditor(queryId, dataUrl) {
         document.getElementById('id_sql').classList.add('changed-input');
     });
     this.bind();
+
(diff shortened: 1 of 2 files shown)


chrisclark:
Another great PR. Cheers!

Issue #214, opened 2016-07-12
UX Enhancement: Allow Users to Expand/Collapse Schemas


raiderrobert:
The following is a little nit picky, but follow me on this. For large databases, you could have lots of schemas, since the search only looks at the schema names, it would make sense to allow for all the fields below to be collapsed by default, so you don't have to scroll for a while.

![screen shot 2016-07-12 at 12 07 37 pm](https://cloud.githubusercontent.com/assets/806571/16774260/3d7f77fe-4829-11e6-8618-21072bd37c0c.png)

chrisclark:
Maybe the thing to do is render anchor links for each table, and sort of have a 'table of contents' at the top of the schema explorer. Clicking on the name of a table jumps you right to that table. I'd prefer something like that to toggling because it doesn't force another click. If someone wants to search for something directly, they can still do that very quickly without data being hidden behind a toggle.

jproffitt:
What do you think about having a "collapse all" button at the top? So they would still all be expanded by default, but you could collapse them all with one click. And also be able to collapse/expand each one individually.

chrisclark:
Works for me. I don't have strong opinions about this feature as long as it doesn't worsen the experience for users who don't have large schemas.

chrisclark:
It suddenly clicked for me that you guys work at the same place :) This is making all sorts of sense now!

raiderrobert:
@chrisclark yeah, I've been super busy, so I asked @jproffitt to take the task from me.

PR #224, opened 2016-07-29
Schema helper improvements

### LINKED ISSUE #223: UX Enhancement: Allow users to auto-insert table and columns names by double clicking in the schema helper
I find myself constantly double clicking to select an item in the schema helper, then copy and pasting it into the SQL editor. It might be nice if you could just double click, and it would automatically insert it into the editor wherever the cursor is.

Do you think this would be a good idea?

jproffitt:
This fixes #223 and #214.

Allows users to expand or collapse each table in the schema helper as well as expand or collapse all of the tables at once.

Allows users to double click on column and table names to automatically insert the column/table name into the sql editor at the cursor location (or replace the currently selected text).

Let me know if you like this, or if I need to change anything.

### DIFF
--- a/explorer/static/explorer/explorer.css
+++ b/explorer/static/explorer/explorer.css
@@ -105,3 +105,27 @@ span.sort {
 #schema_frame {
     width: 100%;
 }
+
+.schema-header {
+    cursor: pointer;
+}
+.insertable {
+    cursor: default;
+}
+
+.CodeMirror-lines .inserted {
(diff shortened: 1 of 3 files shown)


chrisclark:
Looks good. For readability I'd normally pull the actual function bound to the dblclick event into a separate function (so, if we had front end tests, it could be tested separately), but other than that looks great, and that's not enough to keep me from merging :) Thanks!

PR #242, opened 2016-11-01
Dry the permission logic on the views.

FooBarQuaxx:


chrisclark:
Interesting -- I love the result of this (a clean, declarative way of defining permissions), but I don't love the implementation. It feels a bit too meta-y for what is basically a simple CRUD app and makes this bit of code unapproachable to a lot of developers (e.g. me -- I am not smart enough to program with metaclasses).

Also looks like tests are failing because (I suspect) Django changed something about how the class-based views' metaclasses worked as of 1.8, which causes some kind of conflict with the new permission metaclass. The fact that it introduces that level of complexity and surface area is, in my opinion, validation that the juice doesn't seem worth the squeeze here.

But like I said, I love the result -- I just wonder if there is a way to do it that's a little more approachable from a complexity standpoint.

If the solve for getting tests passing is straightforward, I'd reconsider -- but it really feels like a non metaclass-based solution should be possible.

Regardless of where we land on this -- THANK YOU for the thoughtful PR!

chrisclark:
Haven't heard back so I'm going to close this. Happy to reopen the discussion at any time though.

FooBarQuaxx:
Sorry I forget about this. I will try to work it soon. And thanks for your elaborate response by the way.

PR #243, opened 2016-11-01
Use JsonResponse when appropriate.

FooBarQuaxx:


chrisclark:
Makes sense, thanks for contributing!

PR #249, opened 2016-11-06
Add pdf export capability using xhtml2pdf

### LINKED ISSUE #207: [Enhancements] PDF Export
Being able to generate PDF reports would be an excellent addition :)

Cheers, Mykl

spapas:
Add an extra exporter (PDF) using xhtml2pdf and reportlab.  Uses the ``exporer/pdf_template.html`` as a template for the pdf table, you can override it in your own apps to add (for example) page numbers, extra fonts, change page layout etc. Should fix #207.


spapas:
Added a test which passes. The failing travis-ci is because of djcelery, I'm not sure why. I have commented out djcelery references from ``test_project/settings.py`` and everything seems to work.

grantmcconnaughey:
Celery just got upgraded to version 4.0, which moved the `timeutils` module. I don't think we're pinning the Celery version. We'll need to pin it to 3.x until djcelery gets updated to support Celery 4. http://docs.celeryproject.org/en/latest/whatsnew-4.0.html#changes-to-internal-api

spapas:
@grantmcconnaughey cool! However, since this is not really related to the PR I propose to push this change (pinning djcelery version) to master through a different PR and I will then re-merge master to my pdf-export branch.

Thanks!

chrisclark:
@spapas pinned! Feel free rebase from master

spapas:
It seems that django-xhtml2pdf does not support python 3, as per https://github.com/xhtml2pdf/xhtml2pdf/issues/190

How could we proceed with this ?

chrisclark:
I'm OK with the idea of feature-detecting py2 vs. py3 and only executing the tests, and including the exporter, if it's py2. Please just update the readme to reflect that. Sound like the right approach to you?

spapas:
@chrisclark yes that's great thanks! I'll update the PR soon.

explorer/app_settings.py
```diff
@@ -17,7 +18,13 @@
     ('csv', 'explorer.exporters.CSVExporter'),
     ('excel', 'explorer.exporters.ExcelExporter'),
     ('json', 'explorer.exporters.JSONExporter'),
+
 ])
+if sys.version_info[0] < 3:
```
chrisclark: Since this stuff is in the optional requirements, I think we should do a try...except around importing one of the dependencies as well as checking the Python version before importing it. Otherwise we leave open the possibility of runtime errors when folks select the PDF download and the deps aren't installed.

explorer/tests/test_exporters.py
```diff
@@ -75,4 +75,28 @@ def test_writing_excel(self):
 
         expected = b('PK')
 
-        self.assertEqual(res[:2], expected)
\ No newline at end of file
+        self.assertEqual(res[:2], expected)
+
+
+
+@unittest.skipIf(sys.version_info[0] > 2,  "not supported in this library version")
```
chrisclark: I think the move is to probably add a utility function like

```
def pdf_supported():
    # try to import the dependency
    # and python version == 2
    return True else False
```

Then use that both here and in app_settings.py

chrisclark:
Took at a look at the PR. It's generally fine...but it gives me some anxiety that xhtml2pdf is an abandoned project, and the django integration is 4+ years old without any maintenance. The xhtml2pdf project recommends using weasyprint, which looks incredibly simple, has fewer dependencies, and supports Python3. The exporter would just be something like:

```
from django.template.loader import render_to_string
output = BytesIO()
ctx = {
    'headers': res.header_strings,
    'data': res.data,
}
html = weasyprint.HTML(string=render_to_string('explorer/pdf_template.html', ctx)
html.write_pdf(target=output)
return output
```

If you're up for doing it WeasyPrint, I'll merge in a heartbeat. If not, I get it, and I would reluctantly accept this PR since it's clear people need it, but I'd open an issue to replace with WeasyPrint as soon as someone can find the time.

Would love to hear feedback from others as well.

explorer/app_settings.py
spapas: Hello @chrisclark , I've added it in a try catch with the correct import - if python version > 2 or django-xhtml2pdf is not installed pdf exported will not be included.

explorer/tests/test_exporters.py
spapas: I don't really like adding a utility function (in utils.py) because that would introduce a circular import. For the testing if pdf exporting we can just check if python version < 2 (the optional requirements should be installed anyway for testing). So I propose leaving it like this (I've just improved the message a bit).

spapas:
Hello @chrisclark, first of all, xhtml2pdf is not an abandoned project, there is an effort of adding python 3 support to it. Also, the django-xhtml2pdf is just a single view with 10 lines (integrate django templates with xhtml2pdf) that's why it doesn't actually need any updates. 

Both xhtml2pdf  and django-xhtml2pdf work fine with all my (django 1.10) projects.

Now, for weasyprint. I am aware of that project but I don't use it since it is really difficult (if possible at allo) installing it on windows *and* I am using windows as my dev machine. 

Please take a look at its requirements http://weasyprint.readthedocs.io/en/latest/install.html#installing, and more importantly its windows installation: http://weasyprint.readthedocs.io/en/latest/install.html#windows. I refuse to do all this stuff when, using xhtml2pdf is one ``pip install reportlab xhtml2pdf django-xhtml2pdf`` away! 

So I can't go the weasyprint way :( 

What I recommend is merge my PR but, if somebody wants to implement support for WeasyPrint pdf export, implement this *in addition* to xhtml2pdf pdf export, i.e check if weasyprint is installed and if not check if django-xhtml2pdf is installed. That's what I'm going to do at least, I can't use WeasyPrint until it's easy to install on windows *and* xhtml2pdf works really good.

explorer/tests/test_exporters.py
chrisclark: That makes sense, thanks.

chrisclark:
Hah, yep -- that is definitely not a fun install experience. Just to clarify though; it is absolutely an abandoned project. From the home page:

> As per 2016-06-05, this project is officially abandonded by the only maintainer and effectively dead. Please read the notice of 2016-05-02. xhtml2pdf exists by name, it no longer has any relevance.

But like I said, Im ok merging since it's a solid solution. Thanks for all the work on it!

spapas:
Ooh I didn't know that. Really sorry xhtml2pdf is dead ... It served me (and still serves me actually) well in a lot of projects. Thanks for merging though, this will be very useful 👍

spapas:
Hello, just a heads up - xhtml2pdf just got a new maintainer (https://github.com/xhtml2pdf/xhtml2pdf/issues/317) and was resurrected from the dead! So you no longer have to feel guilty for merging this PR 😎

chrisclark:
Phew! I have't slept in 2 weeks!

And seriously - thanks for the update. This does actually reduce my anxiety.

PR #252, opened 2016-11-09
Fix python3 problem with download Excel file

rombr:
Fix python3 problem with download Excel file
(xlsx file downloads as text file, but need binary)


chrisclark:
Looks right to me. Thanks!

PR #254, opened 2016-11-16
Query parameter authentication

noazark:
This PR adds support for authenticating requests with a token in the url:

```
https://example.com/sql-explorer/query/1/?token=hello-world
```

The token is `EXPLORER_TOKEN`, the same as currently used with header based authentication. Query parameter authentication is useful in the cases where clients do not provide access to custom headers.


chrisclark:
Looks great, thanks!

Issue #251, opened 2016-11-07
install and run instruction for none Django users


swuecho:
the readme assume user have Django knowledge. should be more general

jayrav13:
@swuecho I recently used this project without Django experience and totally agree, will put together a README and submit a PR shortly!

PR #260, opened 2016-11-27
Adding simple Django starter instructions in README.

jayrav13:
Title says it all - simple instructions. I realize more can be added, but this is just a start to address Issue #251 .

Jay


chrisclark:
Seems reasonable. Thanks!

PR #264, opened 2016-12-09
Always use latest minor Django version in testing

dsanders11:
Nicked this from another project, it should ensure that Travis is always testing with the latest minor version of Django without having to manually update the file constantly.


chrisclark:
Super smart. Good find!

PR #271, opened 2016-12-28
Db agnostic

startmat:
models.py: changed fetchall to fetchmany with default 1000

schema.py: using django introspection to get schema data

### DIFF
--- a/explorer/models.py
+++ b/explorer/models.py
@@ -1,14 +1,16 @@
 import logging
 from time import time
+
+from django.conf import settings
+from django.core.urlresolvers import reverse
+from django.db import models, DatabaseError, connections
 import six
 
-from django.db import models, DatabaseError, connections
-from django.core.urlresolvers import reverse
-from django.conf import settings
-
-from . import app_settings
 from explorer.utils import (passes_blacklist, swap_params, extract_params, shared_dict_update,
                             get_s3_connection, get_params_for_url)
+
+from . import app_settings
+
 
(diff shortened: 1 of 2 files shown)


chrisclark:
Hey @startmat -- I'm not entirely sure I understand this. What if the user wants > 1000 rows? I think it makes some good sense to perhaps limit the query result to the number of requested rows, but we'd need to plumb this change up through the view level, not just restrict it universally to 1000 rows.

Regarding the change to the schema, this looks like an interesting approach. But is it solving a particular problem or is it just a different way of accomplishing the same thing?

Either way, thanks for taking the time and submitting the PR, I just need some more information about the intent!

startmat:
Hi Chris,
I was testing the sql-explorer with an table which has about 600000 rows. Since there is no paging in the QueryResult I got timeouts in the browser. So I added the limit. I can see two options to solve the problem:

1) Implement paging. This could be painful since it has to be database independent. Oracle paging works different than postgres or mysql. I took a look into the django ORM if it is possible to reuse the paging stuff from there. I did not found an simple way to do that ...

2) Let the user decide how many rows he want to see in the browser.

The change to the schema is using the django ORM introspection, so it is usable for all databases which are supported by django. Your database specific classes are not used in that case. So it is a easier way to provide an tested out of the box solution  for more databases-typs (oracle, mysql, ...)

chrisclark:
I implemented an introspection-based approach to schema in the sb-agnostic branch. I will take another look at the row limiter another time. In the mean time, I'm going to close this PR. Thanks for the inspiration!

PR #275, opened 2017-01-23
Small fix for Google App Engine

Kulikovpavel:
On Google App Engine you get this error, when use Excel Export
Exception Type: NotImplementedError
Exception Value: Only tempfile.TemporaryFile is available for use

With this fix, xlsxwriter use in_memory mode, so all work good


chrisclark:
Sorry this took so long for me to merge - I was put off by the failing tests, but have now looked at them and they seem totally unrelated to the change. Thanks!

PR #280, opened 2017-02-09
update factory-boy dependency version to fix tests

mmautner:
@chrisclark master is failing: https://travis-ci.org/groveco/django-sql-explorer

See https://github.com/FactoryBoy/factory_boy/issues/334


chrisclark:
Thanks!

PR #281, opened 2017-02-13
Retry query snapshot

Anton-Shutik:
Retry query snapshot if failed


Anton-Shutik:
[HTTPError: 500 Server Error: Internal Server Error for url: http://s3.amazonaws.com/ep-reports/query-325.sna...](https://trello.com/c/5TPvRUUG/5681-httperror-500-server-error-internal-server-error-for-url-http-s3-amazonaws-com-ep-reports-query-325-sna)

PR #284, opened 2017-03-06
Remove unused lambda expressions assigned to "_"

JshWright:
This commit removes several noop lambda expressions assigned to the name
"_". In all but one case, they were unused. In the case where it was
being called, the call was simply removed (the lambda was just returing
the value passed to it, so it was unnecessary).

_('foo') is a common idiom for localization, and my guess is that at
some point some code using that convention was copy/pasted in to this
project, and the lambda was added as a workaround for that function
being missing.

### DIFF
--- a/explorer/actions.py
+++ b/explorer/actions.py
@@ -6,8 +6,6 @@ from wsgiref.util import FileWrapper
 from django.http import HttpResponse
 from explorer.exporters import CSVExporter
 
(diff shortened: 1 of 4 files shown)


chrisclark:
Yep! You are correct. That bit of the codebase is super old and definitely unnecessary. Thanks for the contribution!

PR #285, opened 2017-03-07
Fix duplicate test names

JshWright:
The two token auth related tests had the same name. This commit gives
them unique, descriptive names.


chrisclark:
:facepalm: thanks!

Issue #267, opened 2016-12-15
XSS vulnerability


callmewind:
Hi, first of all, thanks for this project; it saved me a lot of time.
I've just discovered that ouput from database is not sanitized berfore rendering it in html, and as far as I know, this makes this application vulnerable to XSS attacks.
It would be nice that EXPLORER_TRANSFORMS could skip the html sanitization, but documentation should clarify that it's use with char fields can be dangerous.

JshWright:
This appears to be an intentional choice (autoescaping is turned off in the two templates that handle rendering of data). I would be curious what the reasoning is for this is... It's a very unsafe default, and certainly violates the "principle of least surprise" mentioned in the README.

chrisclark:
I supposed this is true...the idea is that you can write sql like:

```
SELECT '<a href=http://site.com/user/' || id || '>' || profile_pic || '</a>' FROM users
```

And, for instance, render a gallery of users' profile photos in the results area. But I do see the issue -- it just doesn't come up with the type of data we store in our main explorer target.

I think the fix you proposed @JshWright -- #286 -- makes a lot of sense.

PR #286, opened 2017-03-07
Escape strings from the database when rendering

JshWright:
Disabling autoescaping is unsafe, and results in an XSS vulnerability if
the attacker is able to control database values which get included in
DSE rendered reports.

This commit enables autoescaping by default, and adds a clearly named
setting to restore the old behavior in case someone was relying on it.

This would address #267


chrisclark:
Yep, this makes sense. I left a more in depth comment on #267

PR #291, opened 2017-05-11
Fix #269 - Schema not displaying in MySQL

### LINKED ISSUE #269: Schema Preview does not show anything when database is MySQL
After a bit of digging, the problem is that the app name is hardcoded in the SQL query run
https://github.com/groveco/django-sql-explorer/blob/master/explorer/schema.py#L94

If we make it the app name, it should work.

artschwagerb:
Fix #269 

Uses the current database connection to determine the database name to filter the information_schema.columns table.


ihoru:
@groveco can you merge this PR and release a new version?

chrisclark:
🤦‍♂️ 
Good fix, thanks.

PR #296, opened 2017-07-26
Fixed Postgres Query for Tables

raiderrobert:
The original query assumed that public was always the schema being used; this assumption does not always hold true. Have edited to allow for override via the connection string.


chrisclark:
Thanks @raiderrobert -- this is a good change. However, I'm just about to merge the db-agnostic branch which has a completely new schema builder that should make this PR unnecessary. I'm going to close as a result.

raiderrobert:
@chrisclark great!

PR #299, opened 2017-08-24
(Spell check and) making more robust

drummonds:
Explorer was crashing as I have some decimals stored in it.  This just makes it a bit more robust.  It is not the best possible solution but is useful for me.


chrisclark:
Good catch on the spelling. Got most of the other changes via #312.

Also for future reference, generally don’t bump the version number in the PR; I’ll take care of that during the release process.

Thanks much! I’ll fix the spelling correct directly in master.

PR #300, opened 2017-09-18
Removes invalid excel characters from worksheet name

gosborn:
On Excel export, an exception will be raised by `xlsxwriter` if the worksheet name contains invalid characters. https://github.com/jmcnamara/XlsxWriter/blob/81d1ed9f32887f5c6e615b862fcaf3c8bf49c1c2/xlsxwriter/workbook.py#L720

This PR prevents that exception from being raised by replacing an invalid character with a whitespace. We've seen this be raised most commonly by people who have slashes in the query title name.

```
File "/home/ubuntu/env/local/lib/python2.7/site-packages/explorer/exporters.py" in _get_output
  116.         ws = wb.add_worksheet(name=title)

File "/home/ubuntu/env/local/lib/python2.7/site-packages/xlsxwriter/workbook.py" in add_worksheet
  171.         return self._add_sheet(name, is_chartsheet=False)

File "/home/ubuntu/env/local/lib/python2.7/site-packages/xlsxwriter/workbook.py" in _add_sheet
  635.         name = self._check_sheetname(name, is_chartsheet)

File "/home/ubuntu/env/local/lib/python2.7/site-packages/xlsxwriter/workbook.py" in _check_sheetname
  695.                 sheetname)

Exception Type: Exception at /reports/44/download
Exception Value: Invalid Excel character '[]:*?/\' in sheetname 'Investor Metrics - New/Existing'
```

### DIFF
--- a/explorer/exporters.py
+++ b/explorer/exporters.py
@@ -99,7 +99,6 @@ class ExcelExporter(BaseExporter):
     name = 'Excel'
     content_type = 'application/vnd.ms-excel'
     file_extension = '.xlsx'
-   
 
     def _get_output(self, res, **kwargs):
         import xlsxwriter
--- a/explorer/exporters.py
+++ b/explorer/exporters.py
@@ -99,6 +99,7 @@ class ExcelExporter(BaseExporter):
     name = 'Excel'
     content_type = 'application/vnd.ms-excel'
     file_extension = '.xlsx'
+   
 
     def _get_output(self, res, **kwargs):
         import xlsxwriter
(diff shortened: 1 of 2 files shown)


chrisclark:
This is totally reasonable. But what do you think of using django’s [Slugify](https://docs.djangoproject.com/en/2.0/ref/utils/#module-django.utils.text) function instead of rolling a regex? Super minor, but...I am terrified of regular expressions.

I can take a swing at it as well, though would rather you get the credit!

If neither of us get around to it in the next couple of days, I will just merge as is :) LMK, thanks.

gosborn:
I think using slugify is great! I've updated the PR to use it. From our use, it appears that most reports are titled something like "Attorney Signup Report", so "attorney-signup-report" is very reasonable. In cases where slugify would remove a special character, i.e. "Attorney/Paralegal Signup Report", the formatting is a little less desirable ("attorneyparalegal-signup-report"), but I too am wary of regular expressions - I think using slugify more than makes up for that less desirable formatting (seems to be an edge case anyways) in code-cleanliness and maintenance. Thanks for the suggestion!

chrisclark:
Terrific - thanks!

PR #305, opened 2017-11-11
Transforms are marked safe to enable html

gslaboch:


chrisclark:
Hmmmm, this makes sense on one level, given that the example of how to use transforms in the readme involves rendering html. But on the other hand, transforms can also be used for arbitrary text transformation; and it would be somewhat surprising to get unsafe data back from the server just because you used the transform feature. It also effectively gives *all* users the ability to mark_unsafe *any* text if they know the name of a transform (they can trigger on any column as a result).

I think I’d rather solve this with better documentation for now.

Does that sound right? Open to feedback, or hearing that I’ve misunderstood.

Thanks!

Issue #288, opened 2017-03-20
No longer working on Python 3


jleclanche:
https://github.com/groveco/django-sql-explorer/blob/8168d66050e5bf3fe8f6fcbeb1a9bcb8318eb1a9/explorer/schema.py#L3

This line is breaking everything. It should be `from explorer.app_settings import ...`.

Better yet, all imports should move to relative imports with `from .app_settings import ...`.

I can PR either fixes.

chrisclark:
Yikes. I will fix now.

I will change to `explorer.app_settings` and do a point release in 24hrs (in case any other issues are reported).

Sorry. Any idea why tests didn't catch it? Travis tests against py3.

jleclanche:
Good question; either you're adding the explorer module in the pythonpath (which you shouldn't do), or you're not reaching the schema file from travis.

Thanks for fixing!

Kulikovpavel:
In pip repository still old version 1.1

PR #312, opened 2017-12-11
Fix import for Python 3

daradib:
Similar to #288 but fixes imports in other modules.
Also reorder Travis config to install correct Django version.


chrisclark:
Terrific! Note sure how all of those nasty relative imports snuck their way in in the first place ;)

PR #310, opened 2017-12-05
added django 2.0 compatibility

bradmerlin:
Fixes https://github.com/groveco/django-sql-explorer/issues/309
* Improved import lines to better accommodate new `django.urls` imports.
* Added now-required `on_delete` arguments to migrations so that they match models.
* Reordered travis install order to ensure that correct Django version is always installed.

PR #313, opened 2017-12-11
Django 2.0 compatibility

daradib:
Based on #310, supports `User.is_anonymous` as both a method or attribute (Django 1.10+).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,9 +4,9 @@ python:
   - 3.5
   - 3.6
 install:
+  - pip install ${DJANGO}
   - pip install -r requirements.txt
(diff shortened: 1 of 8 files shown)


daneah:
I mistakenly set out to write a similar pull request this morning without checking if one had already been written. My updates were similar but less thorough and somewhat less robust. Looks great to me!

chrisclark:
I will look through this weekend and try to get merged! Thanks for the work :)

chrisclark:
This is lovely; I will merge now and try to get a release out by the end of the weekend. Thanks very much!

PR #323, opened 2018-03-07
Cast UUID to string, because XlsxWriter can't write it

timmartin:
Apparently this is a deliberate feature of XlsxWriter, because it tries to convert Python types to native Excel types, and there is no native UUID type in Excel. The suggestion is that if you want something mapped to string, you should do it before writing it to XlsxWriter.

### DIFF
--- a/explorer/exporters.py
+++ b/explorer/exporters.py
@@ -1,6 +1,7 @@
 from django.db import DatabaseError
 from django.core.serializers.json import DjangoJSONEncoder
 import json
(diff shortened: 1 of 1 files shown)


chrisclark:
If you can remove the extra migration file, I'm happy to merge. Thanks!

timmartin:
OK, I've removed the extra migration.

chrisclark:
Thanks!

PR #324, opened 2018-03-19
Load https images from cdn instead of http

tarungarg546:
Previously, we were loading only `http` images as background-image for `asc`, `sort` and other actions, while best practice is to load HTTP/HTTPS based on requesters preference.


tarungarg546:
As previously, images were always loading over HTTP, even if your site is fully secure and HTTPS protected, some modern browsers will show that your site is not fully secure and which may lead to doubt in user's mind. 
image

chrisclark:
Yep, very sensible. Thank you!

PR #329, opened 2018-05-16
Fixes

mluard:
The connections select list wasn't populating for me, which broke the url for the schema iframe.  The zip builtin returns an iterator that must have been getting evaluated more than once.  Changing to a list fixed the problem.

Fixing the import in build_schema_cache_async caused a couple of tests to fail because the patches in those tests weren't actually working.  This includes a new test that fails unless the import is fixed, a sanity check in the broken tests to ensure that the patched version of _get_includes is called, and fixes for the patches in test_schema.py

### DIFF
--- a/explorer/forms.py
+++ b/explorer/forms.py
@@ -56,7 +56,7 @@ class QueryForm(ModelForm):
 
     @property
     def connections(self):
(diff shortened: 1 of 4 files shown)


chrisclark:
This is great, thank you!

PR #334, opened 2018-08-13
Updating to include django 2.1

jtrain:


Anton-Shutik:
@jtrain Could you pls resolve conflicts ?

explorer/utils.py
```diff
@@ -2,12 +2,7 @@
 import re
 
 from django.contrib.auth.forms import AuthenticationForm
-
-try:
-    from django.contrib.auth.views import login
-except ImportError:
-    pass
-
+from django.contrib.auth.views import LoginView
```
Anton-Shutik: That breaks backwards compatibility right ?
Can we make the code working on previous django version too?

jtrain: It was introduced django 1.11 earlier versions of django aren't supported by django foundation anymore.

jtrain: https://docs.djangoproject.com/en/1.11/topics/auth/default/#django.contrib.auth.views.login

Anton-Shutik: Cool, can we make tests passing by removing deprecated django versions ?

jtrain:
I have cleaned the documentation/requirements/setup.py file for new supported versions >=1.11,<2.2

We use sql-explorer in Django 2.2 already, but tests fail. I will submit fixes as a separate PR for 2.2 support.

Anton-Shutik:
@jtrain Thanks so much for you contribution! 
So, we should _either_ fix tests and rename PR with "Updating to include django 2.1-2.2" _or_ just drop django 2.2 support and add it in separate PR.

Just making sure we're merging PR that has all tests passing

PR #336, opened 2018-08-16
Fix  auth.views.login (to support Django2.1)

harmv:
Update code to support Django2.1 

django.contrib.auth.views.login was removed from Django 2.1
Use django.conrib.auth.login instead.


harmv:
I'll check the failures -> closing

I'll make a new pull request after fixing

PR #337, opened 2018-08-28
Fix Oracle Query Failure caused by TextField on Group By Clause

ahmetkucuk:
Annotation in `Query` puts `sql` and `description` into Group By clause which fails in Oracle. In Oracle, NCLOB cannot be used in group by clause. See similar issue report:
 
https://code.djangoproject.com/ticket/24096

Seems like there are two solution, either `.defer('sql', 'description')` or using prefetch and count. Former will create several extra queries when we try to reach those attributes. Implemented approach will create 1 extra query to fetch `querylog_set`.

TextField cannot be nullable in Oracle.


marksweb:
@ahmetkucuk I'm just going over old PRs and this is quite old now - is this still relevant?

ahmetkucuk:
I think it is relevant because I still use the library from my own fork because of this issue.

marksweb:
@ahmetkucuk thanks. I don't have any access to oracle so will merge this on the basis that you're running it successfully.

However, could you please merge master back to your fork to update this PR, because there's a 0008 migration in master which looks like it'll conflict with these changes.

ahmetkucuk:
I updated the migrations. I didn't check if there are any new Oracle issues but I will open another PR if there is.

PR #346, opened 2018-11-05
URL-friendly parameter encoding

dmitrins:
Sometimes it is necessary to pass special characters like ' & ' in parameters, e.g.: 
`SELECT * FROM table WHERE field = '$$Param1:p1&p2$$'
`
serializeParams doesn't encode the '&' character properly and it breaks query string parameter pairs.


dmitrins:
I've realized that this fix doesn't need any changes on server side. After removing last commits PR has been automatically closed and Travis CI has failed on the next commit . So I create new one and now it looks good.

PR #349, opened 2018-11-28
Custom Login URL

mmmcorpsvit:
New setting EXPLORER_LOGIN_URL allow set custom login url


chrisclark:
This seems a little out of the blue; isn't there a way in Django to update the default login URL, if someone wanted to use a different view? It seems weird to implement this specifically for Explorer vs. a more general-purpose Django solution.

mmmcorpsvit:
1. i use custom login page protected Google Capthca and some  others features
2. if somebody enter to explorer page so this page will render standart login+password page
3. my PR need for redirect in this case, this is very important because protect site from bruteforce!

Anton-Shutik:
Closing this PR since it's out of scope this repo

PR #353, opened 2019-01-16
Add an unmade migration for the query.connection help text

airhorns:
This was making our `makemigrations --check` on our app fail. I don't think it actually changes anything in the database but Django needs it to be there to not make any new migrations! I think this is from 2 years ago when the help text was added after the field. 

💟


chrisclark:
Just merged an alternative solution for this.

PR #355, opened 2019-02-09
Use configured admin's default site login_template

wgordon17:
If the developer modifies their default admin site `login_template`, this will allow `explorer` to respect the specified `login_template`. This has the benefit of also making no change if the developer does not change the default admin site.

Override default admin: https://docs.djangoproject.com/en/dev/ref/contrib/admin/#overriding-the-default-admin-site
Changing the admin login_template: https://docs.djangoproject.com/en/dev/ref/contrib/admin/#django.contrib.admin.AdminSite.login_template

According do the docs, the `login_template` attribute dates back to 1.7 at least.


meIias:
hi @wgordon17, thanks for your PR. tests seem to be failing here; think you can make the necessary updates to get the build passing before we merge?

PR #359, opened 2019-03-08
Fix unicode issue when generating migration with py2 or py3

MounirMesselmeni:
Running makemigrations on Python 3 will generate a new migration because of Query.connection.help_text which is not unicode on Python 2 which lead to a migration file with byte see explorer/migrations/0006_query_connection.py
```
migrations.AddField(
            model_name='query',
            name='connection',
            field=models.CharField(help_text=b'Name of DB connection (as specified in settings) to use for this query. Will use EXPLORER_DEFAULT_CONNECTION if left blank', max_length=128, null=True, blank=True),
        ),
```
Adding `from __future__ import unicode_literals` will solve this problem but I've to add the right migration any way so that Django will not try generate a new one using Python3.

I have also added minor fixes to use `@six.python_2_unicode_compatible`

I hope this will be merged and released as soon as possible.

### DIFF
--- a/explorer/migrations/0008_auto_20190308_1642.py
+++ b/explorer/migrations/0008_auto_20190308_1642.py
@@ -1,0 +1,19 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('explorer', '0007_querylog_connection'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='query',
+            name='connection',
+            field=models.CharField(blank=True, help_text='Name of DB connection (as specified in settings) to use for this query. Will use EXPLORER_DEFAULT_CONNECTION if left blank', max_length=128, null=True),
+        ),
+    ]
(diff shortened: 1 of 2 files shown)


MounirMesselmeni:
@chrisclark This is basically doing the same as https://github.com/groveco/django-sql-explorer/pull/353 but fixing some unicode compatibility with Python 3 and the reason behind the missing migration is only with Python 3, as in Python 2 there is no migration generated.
Any chance getting this merged and released soon.

chrisclark:
Great change. Approved and merged. I will try to do a release in the next 30 days.

MounirMesselmeni:
Hello @chrisclark , Any chances releasing this soon? :)

saz:
@chrisclark Can you make a new release?

meIias:
Hi all, just wanted to give a heads-up that we are actively working towards a new release and you can expect it soon.

lociii:
Hey @meIias,

any news about the release? We're checking for missing migrations doing a dry-run when deploying our software and have to work around the missing one for this library.
Would be great to be able to get rid off the workaround.

Thanks!

MattFisher:
I don't want to pester, but is the new release going to come out soon? We've got an open PR to do a migration check in CI that's waiting on this fix.
Thanks for an awesome tool!

meIias:
Sorry for the delay folks! the latest version 1.1.3 should be live as of today!

PR #368, opened 2019-08-10
Update EXPLORER_SCHEMA_EXCLUDE_TABLE_PREFIXES

ramonsaraiva:
`EXPLORER_SCHEMA_EXCLUDE_TABLE_PREFIXES` in https://github.com/groveco/django-sql-explorer/blob/master/README.rst are not complying to the code https://github.com/groveco/django-sql-explorer/blob/622b96354e1d7ae7f7105ae90b8da3247e028623/explorer/app_settings.py#L34-L37

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -20,3 +20,4 @@ The following people have contributed to django-sql-explorer:
 - Jared Proffitt
 - Brad Melin
 - Dara Adib
+- Ramon Saraiva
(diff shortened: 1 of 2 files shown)


ramonsaraiva:
I thought about excluding `explorer_` from the defaults as well, but some unit tests broke and seems like they're expecting `explorer_` to be in.. I don't think users would be interested on seeing `explorer_` tables in the schema viewer when writing queries. Maybe in a future PR.

ramonsaraiva:
@chrisclark hey :-)

meIias:
looks reasonable to me. Merging, thanks for the PR @ramonsaraiva!

meIias:
Looks like this should be reverted as it now fails RST file validation:

```
  `long_description` has syntax errors in markup and would not be rendered on PyPI.
    line 368: Error: Unknown target name: "auth".
```

ramonsaraiva:
I don't quite understand that error, @melias

meIias:
Part of the pypi upload process involved validation against the README.rst file, that validation failed when the prefixes were not fully defined like `django.contrib.auth`. I suppose that's why @chrisclark set it the way he did

ramonsaraiva:
`django.contrib.auth` is not a prefix of a table... the documentation is wrong cause it is not excluding those tables from the schemas disaplayed in Explorer.

The table prefixes are in fact `auth_`, etc...

Not sure what process is validating the README file but I don't understand why having `auth_` as a prefix is incorrect.

meIias:
I would try running the README.rst with your changes in it against an RST validator (like this one https://pypi.org/project/rstvalidator/). it throws the same "unknown target name" error.

After a bit of investigation it seems to me like it's just being referenced incorrectly. rather than `auth_` i believe you can do `:ref:auth` and it should work.

PR #369, opened 2019-08-12
Fix ImproperlyConfigured exception

iosifnicolae2:
Fix django.core.exceptions.ImproperlyConfigured: EXPLORER_DEFAULT_CONNECTION is Default, but that alias is not present in the values of EXPLORER_CONNECTIONS


iosifnicolae2:
Too old

PR #370, opened 2019-08-21
Use Django's cookie name setting instead of hardcoded name

MounirMesselmeni:
When some one uses custom csrf cookie name in Django, it needs to override the explorer base.html and js files to make it work.
Default Django CSRF cookie name is csrftoken, it can be changed via settings and if so third party apps should support it out of the box.


Anton-Shutik:
@MounirMesselmeni What's up ? Any description on the PR ?

MounirMesselmeni:
@Anton-Shutik When some one uses custom csrf cookie name in Django, it needs to override the explorer base.html and js files to make it work.
Default Django CSRF cookie name is `csrftoken`

Anton-Shutik:
@MounirMesselmeni Cool, can we add setting description [here](https://github.com/groveco/django-sql-explorer#settings) so people know about that ?

I'd also love to see some tests on that

Thanks for contributing!

marksweb:
@MounirMesselmeni Looking at this, it'd be good to get it merged 👍

chrisclark:
Looks good to me. @marksweb feel free to resolve conflicts and merge! Thanks!

MounirMesselmeni:
@marksweb I updated the PR and added tests. For adding doc about the setting I don't think it's necessary as we will just use the default Django settings for that. Nothing specific for explorer app.

marksweb:
@MounirMesselmeni Excellent work, we do love extra tests!

PR #380, opened 2020-02-09
except Permission denied db exception

songflows:
'ProgrammingError: permission denied for relation table_name' if one of db tables have no access.


marksweb:
@songflows What's the scenario here? How can I replicate this issue?

PR #386, opened 2020-04-24
Django update

Anton-Shutik:
Updated to min supported django version.


Anton-Shutik:
@sramirezgroveco Could you pls take a look ?

Anton-Shutik:
@sramirezgroveco It won't affect our site, since we have the requirements pinned to specific version. And it will not affect us since we're already on django 2.2

PR #393, opened 2020-05-25
Add XlsxWriter module (required to export to Excel) to requirements.txt

maestrow:
Steps to reproduce issue: Try export to Excel any query results.
Actual results: You get an error: ModuleNotFoundError: No module named 'xlsxwriter'.
The solution is to add this module to requirements.txt.


marksweb:
I think I'd prefer this as an optional requirement & not forced upon everyone because it's only for that excel download.

PR #415, opened 2020-10-22
Fix MySQL support: TEXT columns cannot have a default value

natebeacham:
Hey there - just pulled down the latest published version and found out that the recent migrations don't seem to be mysql compatible.

This PR should resolve it be opting for null'ing the column:

Error for reference:

    self = <_mysql.connection open to '127.0.0.1' at 0x5592d51818c0>
    query = b"ALTER TABLE `explorer_query` ALTER COLUMN `description` SET DEFAULT ('')"
        def query(self, query):
            # Since _mysql releases GIL while querying, we need immutable buffer.
            if isinstance(query, bytearray):
                query = bytes(query)
    >       _mysql.connection.query(self, query)
    E       django.db.utils.OperationalError: (1101, "BLOB, TEXT, GEOMETRY or JSON column 'description' can't have a default value")

    venv/lib/python3.8/site-packages/MySQLdb/connections.py:259: OperationalError


marksweb:
Thanks for this @natebeacham 

I think this was introduced by #337 which was for Oracle. It introduced a new migration which I think was similar to what you're talking about.

I've only got Postgres projects at the moment which were happy with it.

I guess that's short lived support for Oracle then!

natebeacham:
@marksweb Hey, thanks for the context. TBH, I don't think either `null=True` or `default=''` are really necessary, and we might get engine parity by omitting both. I'll test out another change.

natebeacham:
Hey @marksweb - I've made the changes, let me know what you think!

marksweb:
@natebeacham Can you merge master into your branch please? I've organised the static files a little better & it's flagging as a conflict.

natebeacham:
@marksweb sure thing - done.

MattFisher:
FYI the previous version of the `0009_auto_20201009_0547` migration failed on our stack with `django.db.utils.OperationalError: cannot ALTER TABLE "explorer_querylog" because it has pending trigger events` using a Postgres 11.5 database that had existing data in the query and querylog tables.
I'm hoping this change will resolve the issue so we can upgrade to v2.

Issue #394, opened 2020-06-23
Failing smoke tests: AttributeError: 'NoneType' object has no attribute 'lower'


jayvdb:
Using `django-smoke-tests`, sql-explorer is breaking on all of its tests.  It doesnt provide the args/params which sql-explorer expects, but that is the point - the forms should describe what they need, and fail form validation sanely if not provided.

The smoke tester has an option to exclude apps, so I have excluded sql-explorer, and as this is a dev-env only tool for us, that is ok, and this isnt a high priority.  Thanks for the great tool ;-)

```py
ERROR: test_smoke_PUT_admin/explorer-ui/new/$ (django_smoke_tests.tests.SmokeTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib/python3.8/site-packages/django_smoke_tests/generator.py", line 90, in test
    response = http_method_function(url, {})
  File "/usr/lib/python3.8/site-packages/django/test/client.py", line 549, in put
    response = super().put(path, data=data, content_type=content_type, secure=secure, **extra)
  File "/usr/lib/python3.8/site-packages/django/test/client.py", line 380, in put
    return self.generic('PUT', path, data, content_type,
  File "/usr/lib/python3.8/site-packages/django/test/client.py", line 421, in generic
    return self.request(**r)
  File "/usr/lib/python3.8/site-packages/django/test/client.py", line 496, in request
    raise exc_value
  File "/usr/lib/python3.8/site-packages/django/core/handlers/exception.py", line 34, in inner
    response = get_response(request)
  File "/usr/lib/python3.8/site-packages/django/core/handlers/base.py", line 115, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/usr/lib/python3.8/site-packages/django/core/handlers/base.py", line 113, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/lib/python3.8/site-packages/django/views/generic/base.py", line 71, in view
    return self.dispatch(request, *args, **kwargs)
  File "/usr/lib/python3.8/site-packages/explorer/views.py", line 89, in dispatch
    return super(PermissionRequiredMixin, self).dispatch(request, *args, **kwargs)
  File "/usr/lib/python3.8/site-packages/django/views/generic/base.py", line 97, in dispatch
    return handler(request, *args, **kwargs)
  File "/usr/lib/python3.8/site-packages/django/views/generic/edit.py", line 149, in put
[...]

PR #425, opened 2020-12-19
models: Require Query.sql

jayvdb:
Fixes https://github.com/groveco/django-sql-explorer/issues/394


marksweb:
Thanks @jayvdb 

That looks like what I had in mind, just a bit busy at the moment!

PR #436, opened 2021-02-13
Fix docs on format of EXPLORER_PERMISSION_VIEW and EXPLORER_PERMISSION_CHANGE

ryanwilsonperkin:
EXPLORER_PERMISSION_VIEW and EXPLORER_PERMISSION_CHANGE examples were
both written as `lambda u: ...` as though the argument passed to them
were a user, but the argument that gets passed is actually a request.

This updates them to match the format that is default in the code.


marksweb:
Thanks, thought I'd done this when the code changed 🤦‍♂️

marksweb (approved):

PR #441, opened 2021-05-14
Fix #430 - CSV exports to include a BOM

### LINKED ISSUE #430: Exports need to include a BOM
Currently exporting a CSV and opening it in Excel ruins any unicode characters.

For example, Japanese characters display as `Ë£ïÂ§™`

Exporter needs to do something like this to write the BOM to the beginning of the file;

```python
    contents = StringIO()
    contents.write(codecs.BOM_UTF8.decode('utf-8'))
    writer = csv.writer(contents)
```

bdista:
- Fix #430 - CSV exports to include a BOM
BOM_UTF8 added via the `CSVExporter` class.
A few existing tests rely on this class and break when adding a BOM. This PR makes them independent of whether a BOM is there or not, and adds a dedicated test for the BOM. Hope that approach is ok.
Test suite passes and using the `/test_project` the Japanese character ガ displays properly when opening the csv in excel.

- Fix paths of requirement files in `/test_project/start.sh`

### DIFF
--- a/explorer/exporters.py
+++ b/explorer/exporters.py
@@ -1,3 +1,4 @@
+import codecs
 import csv
 import json
 import string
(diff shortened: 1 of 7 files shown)


bdista:
Thanks for merging it!

PR #444, opened 2021-06-01
Restructured The Settings Docs

### LINKED ISSUE #443: Re-write the settings docs
The table that's in use on the settings docs doesn't read very nicely/easily given the length of settings names and the amount of information that corresponds with each one.

https://django-sql-explorer.readthedocs.io/en/latest/settings.html

This could look, and read, much better if it was displayed differently, perhaps in sections for each setting.

Patil2099:
Fixes #443

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -1,30 +1,225 @@
+********
 Settings
-========
+********
 
-======================================= =============================================================================================================== ================================================================================================================================================
-Setting                                 Description                                                                                                                                                  Default
-======================================= =============================================================================================================== ================================================================================================================================================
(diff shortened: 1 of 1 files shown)


marksweb:
Hi @Patil2099 - Thank you for looking at this so quickly!

I've got a bit of feedback for the formatting which I'll add.

Also the docs build github action is failing, but I suspect this is because I've changed the docs config since I built that action so maybe the action needs to install the package as well as the requirements.

marksweb (changes requested):

marksweb (approved):

PR #448, opened 2021-07-10
Update utils.py

jwestfall:
Allow parameter values to contain a : and not get dropped.


marksweb:
@jwestfall thank you for your contribution. Could you please add tests for this?

PR #449, opened 2021-07-14
Translates expression for duration

jtrain:
I am running django 3.2 and noticed that the execution time is blank. e.g.

image

https://docs.djangoproject.com/en/2.2/topics/i18n/translation/#blocktrans-template-tag
The django docs are clear that expressions should be bound to a local variable before being used inside the translation tag.

I've chosen to use the ~older format for the expression binding since this should cover a wider range of django installs~ newer format because it is also available for a wide range of django installs.

image

### DIFF
--- a/explorer/templates/explorer/preview_pane.html
+++ b/explorer/templates/explorer/preview_pane.html
@@ -16,7 +16,7 @@
                               <div class="row">
                                 <div class="col-md-6">
                                   {% if data %}<a title="Show row numbers" id="counter-toggle" href="#">#</a>&nbsp;{% endif %}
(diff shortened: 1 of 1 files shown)


marksweb:
@jtrain Thank you for your contribution.

Is this basically django 2.2+ compatible?

jtrain:
I haven't had to make any other changes on django 3.2, its working very nicely.

If you are asking if this specific change is compatible with 2.2+ then yes, in fact it is 1.8+ compatible.

jtrain:
Here are the docs for 1.8 blocktrans https://docs.djangoproject.com/en/1.8/topics/i18n/translation/#blocktrans-template-tag, I used this syntax.

marksweb:
Fantastic - Thank you @jtrain!

PR #466, opened 2022-01-17
Use Django built-in get_valid_filename

SuJiKiNen:
This is more loosen restriction which allowing Unicode in filename.

### DIFF
--- a/explorer/exporters.py
+++ b/explorer/exporters.py
@@ -8,7 +8,7 @@ from io import StringIO, BytesIO
 
 from django.core.serializers.json import DjangoJSONEncoder
 from django.utils.module_loading import import_string
-from django.utils.text import slugify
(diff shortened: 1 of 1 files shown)


marksweb:
@SuJiKiNen Thank you for the contribution. Unfortunately this doesn't pass the tests.

```
FAIL: test_downloading_from_playground (explorer.tests.test_views.TestCSVFromSQL)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/runner/work/django-sql-explorer/django-sql-explorer/explorer/tests/test_views.py", line 506, in test_downloading_from_playground
    resp['Content-Disposition']
AssertionError: 'filename="Playground-1.csv"' not found in 'attachment; filename="Playground-1"'
```

The filename is tested in the view tests and as you can see, it's not returning the result expected.

PR #471, opened 2022-02-12
fix extra space around sql

fearless0307:


marksweb:
What's the actual fix here @fearless0307?

It looks like the markup is the same, just with added line breaks 😂

fearless0307:
@marksweb 
if we use <p> tag, it ignore surrounded space, but if we use <pre> tag, it also consider the surrounded space.
Same issue was here, <textarea> tag was taking surrounded space as value in codemirror editor.

marksweb:
@fearless0307 ah yes I thought that might be why the formatting alone fixed it 👍

Great work. As always I assumed it more complicated & thought it was a problem in the JS!

marksweb:
Released in 2.4.1

https://pypi.org/project/django-sql-explorer/

Issue #483, opened 2022-08-29
Creates bad migration when using DEFAULT_AUTO_FIELD


spapas:
Hello, I'm using the DEFAULT_AUTO_FIELD = "django.db.models.BigAutoField" setting in a project.

When I install django-sql-explorer and run make migrations it will create a bad migration file on the django-sql-explorer instalaltion directory that alters the id fields. It's contents are similar to this:

```
# Generated by Django 4.1 on 2022-08-29 08:44

from django.db import migrations, models


class Migration(migrations.Migration):

    dependencies = [
        ("explorer", "0010_sql_required"),
    ]

    operations = [
        migrations.AlterField(
            model_name="query",
            name="id",
            field=models.BigAutoField(
                auto_created=True, primary_key=True, serialize=False, verbose_name="ID"
            ),
        ),
        migrations.AlterField(
            model_name="querylog",
            name="id",
            field=models.BigAutoField(
                auto_created=True, primary_key=True, serialize=False, verbose_name="ID"
            ),
        ),
    ]


```

To fix that we need to add the line `default_auto_field = django.db.models.AutoField` on apps.py so as to *not* use the `django.db.models.BigAutoField` I'm using on my settings. See  https://docs.djangoproject.com/en/4.1/ref/applications/#django.apps.AppConfig.default_auto_field for reference.

This is a single line change, I'd be happy to provide a PR if needed.

BR,
Serafeim

PR #484, opened 2022-08-29
Fix for #483

spapas:
I actually had to create the PR because I needed to use that instead of the pypi release on my requirements to avoid the bad migration.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -8,6 +8,8 @@ This project adheres to `Semantic Versioning <https://semver.org/>`_.
 `unreleased`_ changes
 ---------------------
 
+* #484: Avoid creating migration when using `BigAutoField` as `DEFAULT_AUTO_FIELD`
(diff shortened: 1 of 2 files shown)


marksweb (approved):

marksweb:
Thanks @spapas, I thought this was in already 🤦

marksweb:
@spapas could you please add this to the unreleased section in the history file please?

spapas:
@marksweb done (sorry but I couldn't make the link to the PR work)

marksweb:
> @marksweb done (sorry but I couldn't make the link to the PR work)

No problem - I can do that. For reference, the link locations are all at the end of the file and then they just get the syntax added to get them to use that location.

spapas:
Ah!!! I thougth that this was some github magic... Yes I only now noticed the .rst extension :) Thank you!

Issue #482, opened 2022-08-26
Save Parameter key and values


dwmorrison33:
I am curious if there is a way to save a query parameters value for later use instead of having to re-enter it each time one goes to the `Query Detail` page? I only ask because I have not been able to find an answer to that with Google and I do not see anything in the codebase to do it, but maybe I am missing something. Thank you

marksweb:
@dwmorrison33 could you provide an example of what you mean please?

dwmorrison33:
@marksweb Thank you for your response, below is an example. I create a new query with `id` as the parameter and execute it:
<img width="1590" alt="Screen Shot 2022-08-27 at 3 46 31 PM" src="https://user-images.githubusercontent.com/13126248/187045833-0f0718a3-3e21-4265-b82a-c5d347cb2da4.png">

After navigating back to the Query List page and clicking again on this query, the parameter `id` no longer has the value `10000` populated:
<img width="1586" alt="Screen Shot 2022-08-27 at 3 46 47 PM" src="https://user-images.githubusercontent.com/13126248/187045879-dab171e5-e8ee-4f16-8e7e-e4e25043fecf.png">

Is there a way to persist this value? Also, on a side note, I am using version `1.1.3` of your package since my company is on Django 1.11 and that is the last version that is compatible.

marksweb:
@dwmorrison33 Right, I'm with you - and I thought that was what you meant. I'm fairly sure the app can't do that, at least as it is now.

My suggestion would be to create bookmarks with the query params defined in them.

dwmorrison33:
> @dwmorrison33 Right, I'm with you - and I thought that was what you meant. I'm fairly sure the app can't do that, at least as it is now.
> 
> My suggestion would be to create bookmarks with the query params defined in them.

Would you mind if I implemented a feature to save the query Params so that we can save a default value and then put in a pull request, as saving bookmarks will become a bit laborious across the organizations users who are using these queries each day. I just want to check before I work on this feature and you all not wanting it, but I think it would be useful to others as well.

marksweb:
@dwmorrison33 Yeah that sounds like it'd be really useful.

Create a branch from the 1.1.3 tag and when you come to a PR, target this new branch I've setup; https://github.com/groveco/django-sql-explorer/tree/release/1.1.4

Then I can do a 1.1.4 release and then, hopefully, merge it to the main branch for it to benefit the users of the latest version as well 👍

dwmorrison33:
[...]

PR #487, opened 2022-09-01
Sql explorer save query params

dwmorrison33:
### Description

This provides the functionality to save query params so that the queries execute with them each time one comes to the Query Detail page.

Resolves: #482

### Deployment Procedures

* [ ] Migration (Please specify app(s) below)
    - `./manage.py migrate explorer`

### Tests

Execute Manual UI Tests like the following:
1. Create a new query with a query param and save, note the same functionality exists as previously where the query does not execute due to no value for the new param:
Screen Shot 2022-09-01 at 10 57 38 AM

2. Input a value for the parameter and save:
Screen Shot 2022-09-01 at 10 57 52 AM

3. Now add a new parameter and save:
Screen Shot 2022-09-01 at 10 58 39 AM

4. Now save a new value to the second parameter:
Screen Shot 2022-09-01 at 10 58 53 AM

5. Now navigate back to the Query List page and click on the test query:
Screen Shot 2022-09-01 at 10 59 08 AM

6. Validate that the new `QueryParams` persist and the query executed:

Screen Shot 2022-09-01 at 10 59 41 AM

Also, you will want to verify that the csv and JSON exports still work correctly like:

Screen Shot 2022-09-01 at 11 08 54 AM

Screen Shot 2022-09-01 at 11 09 13 AM

### DIFF
--- a/explorer/migrations/0009_auto_20220830_1545.py
+++ b/explorer/migrations/0009_auto_20220830_1545.py
@@ -1,0 +1,28 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('explorer', '0008_auto_20190308_1642'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='QueryParam',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('key', models.CharField(blank=True, max_length=128, null=True)),
+                ('value', models.CharField(blank=True, max_length=255, null=True)),
+                ('query', models.ForeignKey(blank=True, null=True, on_delete=django.db.models.deletion.CASCADE, related_name='query_params', to='explorer.Query')),
+            ],
+        ),
+        migrations.AlterUniqueTogether(
+            name='queryparam',
+            unique_together=set([('query', 'key')]),
+        ),
+    ]
--- a/explorer/models.py
+++ b/explorer/models.py
@@ -128,6 +128,21 @@ class Query(models.Model):
             return [SnapShot(k.generate_url(expires_in=0, query_auth=False),
                              k.last_modified) for k in keys_s]
 
+    def create_query_params(self, data):
+        '''
+        Method to create the QueryParam objects and associate them to a Query obj
+        Args:
+            - `data`: dict
+        '''
+        # First remove all associated QueryParam objs before creating new ones
+        self.query_params.all().delete()
+        query_params_to_create = [
+            QueryParam(query=self, key=k, value=v)
+            for k, v in data.items()
+        ]
+        # Now bulk create new ones
+        QueryParam.objects.bulk_create(query_params_to_create)
+
 
 class SnapShot(object):
 
--- a/explorer/models.py
+++ b/explorer/models.py
@@ -41,7 +41,6 @@ class Query(models.Model):
                                   help_text="Name of DB connection (as specified in settings) to use for this query. Will use EXPLORER_DEFAULT_CONNECTION if left blank")
 
     def __init__(self, *args, **kwargs):
-        self.params = kwargs.get('params')
         kwargs.pop('params', None)
         super(Query, self).__init__(*args, **kwargs)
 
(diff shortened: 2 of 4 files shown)


dwmorrison33:
@marksweb Thank you for the instructions here: https://github.com/groveco/django-sql-explorer/issues/482#issuecomment-1233269126 They were very helpful and I learned a couple new things.

marksweb:
@dwmorrison33 You're welcome. I'll take a look over this.

First thing that I notice is the inclusion of `unicode_literals`. Are you using python 2 still? If you are, you should move to python 3.

dwmorrison33:
> @dwmorrison33 You're welcome. I'll take a look over this.
> 
> First thing that I notice is the inclusion of `unicode_literals`. Are you using python 2 still? If you are, you should move to python 3.

@marksweb Im not sure why that was auto generated since I am on Python3 and these code changes were tested against python3.7 and django1.11

marksweb:
@dwmorrison33 So just trying to understand this, because as I said, I don't use query params in my queries.

If a query parameter ends up as a database object linked to the query - why not just put the value into the query and have different queries for however many values you need? Because I'm just thinking about efficiency. If there is value in having the query parameters in the database, I'd maybe have done a JSON field so all the params with a query go into 1 object.

dwmorrison33:
> If a query parameter ends up as a database object linked to the query - why not just put the value into the query and have different queries for however many values you need? 

I just realized that I did not respond to the first question you asked and the reason why we wouldnt want that is because we have roughly 1000 queries and that could become cumbersome to save a bunch of the same queries with different values. The desired behavior is that we can store the query parameters for users who are not experienced with writing SQL and would not have to update values directly, as well as so that after one navigates away from the Query Detail and then comes back that the params would persist for the next user who is looking at that Query Detail.

explorer/migrations/0009_auto_20220830_1545.py
```diff
@@ -0,0 +1,28 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
```
marksweb, line 2: Could you remove these first two lines please? So it starts with the django imports.

explorer/migrations/0009_auto_20220830_1545.py
```diff
@@ -0,0 +1,28 @@
...
+
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('explorer', '0008_auto_20190308_1642'),
```
marksweb, line 11:
```suggestion
        ('explorer', '__latest__'),
```

This migration is the difficult part in porting this to the main branch because it adds a migration in to the past. 

My current thinking is that by depending on the latest migration, this commit will work on this 1.x release or on the current release.

There is a chance you'll not be able to upgrade from this new version, but that might be acceptable to you & we could still release changes on this branched code.

marksweb (changes requested):

explorer/migrations/0009_auto_20220830_1545.py
dwmorrison33: > My current thinking is that by depending on the latest migration, this commit will work on this 1.x release or on the current release.

Good point on the migrations, that was something I had not considered. I will need to confirm with my manager, but Im thinking that we will want these changes to be a part of the current release and not `1.x` release due to the migration challenges as we will be upgrading from Django 1.11 to Django 2.2 in the near future and will also want to upgrade `django-sql-explorer` at that time. I suppose I can import the forked commit into our project and then after we upgrade to Django 2.2, I can then upgrade to the current release. I will of course have to migrate the data at that time, but that is pretty straightforward to do.

All that being said, I will get back to you to confirm that we want to go forward on the current release.

dwmorrison33: @marksweb I talked to my manager and he thinks it best to make this a part of the current release and not `1.x`. That being said, I looked at the [current code](https://github.com/groveco/django-sql-explorer) and I dont think this code will auto merge into that due to the [views being separated](https://github.com/groveco/django-sql-explorer/tree/master/explorer/views) instead of in a single `views.py` file like it is in `1.1.3` release. 

So Im thinking it best to do the changes based off the master branch and then submit a new PR. What are your thoughts on that?

marksweb: Ok this makes sense, branch from master and you can use this PR as reference to bring the code into that branch & then create the migration again once things are in place 👍

dwmorrison33:
Closing in favor of https://github.com/groveco/django-sql-explorer/pull/489

PR #489, opened 2022-09-06
Sql explorer save query params

dwmorrison33:
### Description

This provides the functionality to save query params so that the queries execute with them each time one comes to the Query Detail page.

Resolves: #482

### Deployment Procedures

* [ ] Migration (Please specify app(s) below)
    - `./manage.py migrate explorer`

### Tests

Execute Manual UI Tests like the following:
1. Create a new query with a query param and save, note the same functionality exists as previously where the query does not execute due to no value for the new param:

Screen Shot 2022-09-06 at 2 57 16 PM

2. Input a value for the parameter and save:
Screen Shot 2022-09-06 at 2 57 35 PM

3. Now add a new parameter and save:

Screen Shot 2022-09-06 at 2 58 27 PM

4. Now save a new value to the second parameter:

Screen Shot 2022-09-06 at 2 58 47 PM

5. Now navigate back to the Query List page and click on the test query:

Screen Shot 2022-09-06 at 2 59 10 PM

6. Validate that the new `QueryParams` persist and the query executed:

Screen Shot 2022-09-06 at 2 59 35 PM

Also, you will want to verify that the csv and JSON exports still work correctly like:

Screen Shot 2022-09-06 at 3 04 05 PM

Screen Shot 2022-09-06 at 3 04 22 PM

### DIFF
--- a/explorer/migrations/0011_auto_20220906_0744.py
+++ b/explorer/migrations/0011_auto_20220906_0744.py
@@ -1,0 +1,25 @@
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('explorer', '0010_sql_required'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='QueryParam',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('key', models.CharField(blank=True, max_length=128, null=True)),
+                ('value', models.CharField(blank=True, max_length=255, null=True)),
+                ('query', models.ForeignKey(blank=True, null=True, on_delete=django.db.models.deletion.CASCADE, related_name='query_params', to='explorer.Query')),
+            ],
+        ),
+        migrations.AlterUniqueTogether(
+            name='queryparam',
+            unique_together=set([('query', 'key')]),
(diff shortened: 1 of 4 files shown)


dwmorrison33:
Note: This is tested with Python3.8 and Django3.2

dwmorrison33:
It looks like I need to update some tests and there is one change I think will be best when creating/deleting the query parameters. Will  try to get to those today.

marksweb:
@dwmorrison33 yeah I've not had time to look at the tests but think I saw a change to the way the params attribute works in the view

chrisclark:
Whoa whoa whoa -- there actually IS a way to do this! The param syntax supports providing a default value. As per the docs:

You can combine both a default and label to your parameter but you must start with the label: $$paramName|label:defaultValue$$.

Does that hit the bid?

dwmorrison33:
> Whoa whoa whoa -- there actually IS a way to do this! The param syntax supports providing a default value. As per the docs:
> 
> You can combine both a default and label to your parameter but you must start with the label: paramName|label:defaultValue.
> 
> Does that hit the bid?

@chrisclark Thank you for the info, but this PR is stale. I’m thankful we didn’t move forward with this request last year, I will close this PR now.

chrisclark:
All good! I really appreciate you taking the time and sorry it took uh...a year to weigh in.

Issue #492, opened 2022-09-22
Format button not working


avi-perl:
Clicking the format button is not working. 
When I click the text on the button, nothing happens at all. If I select another part of the button, I get the following:
> Forbidden (403)
> CSRF verification failed. Request aborted.

lawson89:
Ok I just ran into this exact issue on a project of mine and finally figured it out. It was this setting which security had mandated

CSRF_COOKIE_HTTPONLY = True

Explanation from Django docs below

CSRF_COOKIE_HTTPONLY[¶](https://docs.djangoproject.com/en/4.1/ref/settings/#csrf-cookie-httponly)

Default: False

Whether to use HttpOnly flag on the CSRF cookie. If this is set to True, client-side JavaScript will not be able to access the CSRF cookie.

marksweb:
@lawson89 Thank you for clarifying this.

I don't turn that setting on because of the problems it causes in javascript so I've not seen this.

Perhaps we need some additional docs to clarify this django setting in relation to this functionality.

lawson89:
Ok sure I am happy to take a crack at adding this in the docs if you would like

marksweb:
@lawson89 That would be great if you could please! 🥇

PR #496, opened 2022-10-10
Document breakage of Format button due to CSRF_COOKIE_HTTPONLY (#492)

lawson89:
Document breakage of Format button if settings.CSRF_COOKIE_HTTPONLY = True


marksweb (approved):
That's great. Thanks @lawson89

Issue #273, opened 2016-12-31
Avoid execution of parameterised queries when opening


anirvan-majumdar:
I have created and saved a bunch of queries and most of them rely on parameterisation. However, SQL Explorer attempts to execute the query everytime I click and open the query from the query listing page. If I've provided a default parameter value, then the explorer attempts to execute the query with the default, and if I haven't provided a default, it attempts to execute the query with the value set to `''`. In either case, the experience gets very sluggish, since the query execution with the default or '' value can take a _long_, _long_ time to execute.

It would be ideal if parameterised queries are _not_ executed when clicked to open.  

On a similar note, I also felt that there should be a simple "Save" option for queries. I can also see that #227 did indeed address this issue - but I'm not sure whether this has been released as part of version `1.0` which is currently available. Can you let me know when the next release is scheduled?

Thanks!

lawson89:
I'd be happy to take a shot at this as it bothers me as well

@marksweb let me know. Also should this be configurable in settings?

marksweb:
@lawson89 yeah I like the sound of a setting which controls this 👍

Thanks for your help

PR #497, opened 2022-10-11
Avoid execution of parameterised queries when viewing query

lawson89:
Add setting which prevents autorun of parameterized queries

Issue #273


lawson89:
Let me add a test case for the new setting and then I'll update this PR

marksweb (approved):

Issue #454, opened 2021-08-11
Bug in passes_blacklist


afedosenko:
Hello!
I am running the following query:

`SELECT * from eventtype where eventtype.value = 'Grant Date';`

I get:

`Query failed the SQL blacklist: GRANT`

In order for the query to work, I have to do workarounds:
`SELECT * from legal_eventtype where legal_eventtype.value = concat('G', 'rant Date');`
This causes a lot of headaches.

afedosenko:
Hello @marksweb, django-sql-explorer==2.3

marksweb:
Thanks @afedosenko. I was afraid you'd say that 😂 

In `2.1.0` the blacklist checking was significantly improved so I thought it was foolproof by now.

That release fixed #371 and #412

lawson89:
@marksweb Verified this as well. 
Wonder if instead of a regex based approach, using sqlparse would be better since it is already a dependency of the project?
I'd be happy to work on this or fix the current regex problem with blacklisting sql parameters - whichever direction the team prefers

lawson89:
@marksweb I'll start on this one this weekend

marksweb:
@lawson89 Based on the fact it's been established that the current code doesn't work I'd be in favour of removing it and only using something that works.

I'll release it under a major version because it's a significant part of the application. And given the fact we have the current broken version, I imagine the test suite doesn't sufficiently cover it. So if you could also write some tests for this, that would be magnificent.

lawson89:
@marksweb ok will do. There are some test cases but I will add some more - for instance the above case

lawson89:
@marksweb Let me know if you have any questions on this

PR #498, opened 2022-10-18
Change sql blacklist functionality from regex to sqlparse

lawson89:
Issue #454

### DIFF
--- a/explorer/tests/test_utils.py
+++ b/explorer/tests/test_utils.py
@@ -4,7 +4,6 @@ from unittest.mock import Mock
 from django.test import TestCase
 
 from explorer import app_settings
-from explorer.actions import generate_report_action
 from explorer.tests.factories import SimpleQueryFactory
 from explorer.utils import (
     passes_blacklist, param, swap_params, extract_params,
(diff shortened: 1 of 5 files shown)


marksweb (approved):

explorer/tests/test_utils.py
```diff
@@ -17,46 +16,173 @@ class TestSqlBlacklist(TestCase):
...
+        passes, words = passes_blacklist(sql)
+        self.assertFalse(passes)
+
+    def test_dml_rollback(self):
+        sql = "ROLLBACK TO SAVEPOINT my_savepoint;"
+        passes, words = passes_blacklist(sql)
+        self.assertFalse(passes)
+
+    def test_dml_set(self):
+        sql = "SET TIME ZONE 'PST8PDT';"
```
covracer, line 114: @marksweb why forbid this? What's a better way to display results in a local timezone?

marksweb: Forbid this? Sorry I don't follow - I haven't forbidden this?

covracer: This PR added `SET` to `app_settings.EXPLORER_SQL_BLACKLIST`, and the assert below seems to be checking that `SET TIME ZONE 'PST8PDT'` fails the blacklist. I noticed this because several queries I rely on started failing after an upgrade.

From a quick reading of https://www.postgresql.org/docs/current/sql-set.html, I don't notice ways in which it'd modify data in the database, but I'm not a SQL expert, and I haven't checked MySQL, etc.

marksweb: @covracer This PR just changed the order of the blacklisted items. `SET` was actually added about a year ago [here](https://github.com/groveco/django-sql-explorer/commit/ca31cc946de68bc7eb9a5b03601134a70e4b1564).

`SET` can be used to update values. See [here](https://www.w3schools.com/SQL/sql_ref_set.asp).

The blacklist check didn't work for some time. It may have been this PR which actually got it working correctly hence you noticing it.

covracer: https://www.w3schools.com/SQL/sql_ref_set.asp discusses `SET` in the context of `UPDATE ... SET ... WHERE`. It seems to me like the entry for `UPDATE` would cover that.

`SET PASSWORD` for old versions of MySQL as described in #475 sounds like a great thing to block. Would it be possible to disallow `SET PASSWORD ...` while allowing `SET TIME ...`?

Issue #389, opened 2020-05-08
Hitting the Format Button results in a 403 or 405 error


zurtri:
When hitting the format button I get

In Browser:
POST http://localhost:8000/explorer/format/ 403 (Forbidden)

In CLI:
Forbidden (CSRF token missing or incorrect.): /explorer/format/
WARNING 2020-05-08 11:38:33,843 log 9560 1008 Forbidden (CSRF token missing or incorrect.): /explorer/format/
 "POST /explorer/format/ HTTP/1.1" 405 2513

I've refreshed and the checked that the CSRF token is being rendered (it is), but still get the same issue. Saving a query is fine.

Any ideas?

alxnik:
Same issue here albeit in the /explorer/[id]/email_csv endpoint.

It is related to the CSRF_COOKIE_HTTPONLY django setting. If it is enabled, the buttons do not work. Keep in mind that even if you disable the setting, you have to reset the cookie on the browser too (either restart the browser or go to developer tools->storage->find the csrftoken cookie->HttpOnly=False - this is no firefox)

marksweb:
Project setting/config related, not an issue, closing.

PR #533, opened 2023-03-07
CSRF token httponly support + s3 destination for async results

YakiYaki:
Hi!

If `CSRF_COOKIE_HTTPONLY=True`, then [email sending feature won't work](https://github.com/groveco/django-sql-explorer/issues/389#issuecomment-655511516). 
We need this setting to be True, so ask you to consider some adds:

- added support for `CSRF_COOKIE_HTTPONLY=True`: getting csrf token value from hidden input [[docs](https://docs.djangoproject.com/en/4.1/howto/csrf/#acquiring-the-token-if-csrf-use-sessions-or-csrf-cookie-httponly-is-true)]. Only on query list view, where email_csv request is made.
- added app setting `EXPLORER_S3_DESTINATION` for specify destination for async query results. Tried to use `EXPLORER_S3_S3_ENDPOINT_URL` and `EXPLORER_S3_BUCKET`, but no luck.

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -252,6 +252,25 @@ Useful to set if you are using a non-AWS S3 service or you are using a private A
 
 
 
+S3 destination path
+********************
+
+S3 destination path. Defaults to empty string.
+Useful to set destination folder relative to S3 bucket.
+Along with settings ``EXPLORER_S3_ENDPOINT_URL`` and ``EXPLORER_S3_BUCKET`` you can specify full destination path for async query results.
+
+.. code-block:: python
+
+    EXPLORER_S3_DESTINATION = 'explorer/query'
+
+    # if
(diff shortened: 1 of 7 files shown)


marksweb:
Thanks @YakiYaki this looks good.

Could you add an entry to the settings docs though please to explain what this new setting does; https://github.com/groveco/django-sql-explorer/blob/master/docs/settings.rst

And this area of the app doesn't have any tests as far as I'm aware unfortunately - so I'll trust that because you're using this that it is functioning as it should.

marksweb (approved):
Thank you for this @YakiYaki

lucas-giaco:
Hi Folks, by any chance do you have an ETA for a new release including this feature?

marksweb:
@lucas-giaco Just released 3.2 - https://pypi.org/project/django-sql-explorer/

PR #538, opened 2023-05-22
Updated necessary default branch refs

brigitamartin:


marksweb:
`master` is the default branch name.

https://github.com/groveco/django-sql-explorer/branches

PR #539, opened 2023-05-26
Test for SET PASSWORD

covracer:
#475 added SET to the deny list because of how `SET PASSWORD` works in MySQL. #498 added a test for SET, using `SET TIME ZONE` from PostgreSQL. As far as I can tell, denying the session-specific `SET TIME ZONE` will not improve security. So test for `SET PASSWORD` instead.


marksweb:
Yeah, that makes sense! Thanks @covracer

Issue #542, opened 2023-07-08
Indicating whether the query still runs after changes in the DB


eeriksp:
## Problem statement

One issue that keeps happening for me quite often is that the saved queries no longer run after new migrations have been made and the table structure has changed.

It is time-consuming to go over each query manually to check if it still works after a database change has occurred.

## Proposed solution

I propose to add a boolean field `compiles` to the `Query` model which would indicate if the query returns an error or not. This filed would be overridden every time the query is run. In the list view of the queries that could be displayed as a green/red circle, just like boolean values are displayed in Django admin.

There could also be a management command (maybe call it `test_sql_queries`) which would run all queries and thus update their `compiles`  fields. `./manage.py test_sql_queries` could be run after running migrations to identify the broken queries.

Since SQL explorer also permits queries that modify data, there needs to be another boolean flag which allows the query to be excluded from the set of queries  `./manage.py test_sql_queries` runs to prevent any unwanted data modifications.

How does that idea sound?
I could prepare a pull request if that seems to be a reasonable solution.

marksweb:
I'm having a think through this one.

First thing that strikes me though is what that field would get called. A query isn't compiled, it's ran or executed. So I'd be tempted to call the field something like `can_run` to show that the query can be ran.

I'm not sure about a 2nd field though. I think excluding queries and concern for queries modifying data would be better mitigated against using management command arguments. Maybe there could be an "exclude" argument that would exclude specified queries, and a "connection" argument that meant you could ensure the queries are ran on a connection with read-only access ([docs](https://django-sql-explorer.readthedocs.io/en/latest/settings.html#database-connections))

eeriksp:
Yes, your suggestions make sense.
I start to work on the PR.

chrisclark:
Addressed in https://github.com/chrisclark/django-sql-explorer/pull/562

Thanks so much!

PR #545, opened 2023-07-13
WIP:  Indicating whether the query still runs after changes in the DB

eeriksp:
Solves #542


eeriksp:
Here is a screenshot of the list view (notice the "Ran Successfully" column):
image

marksweb:
I'll have a look over this, but there's some test failures & linting issues that look like a good place to start on more changes.

chrisclark:
Heyo! I'm jumping back into some dev here -- I think this is a totally cool idea. I wonder if it is possible to source it from the querylog however, rather than creating a new field and having to manage the state of it. I'll make a stab at it and tell me what you think.

chrisclark:
Here it is, but doesn't require adding any statefulness -- just expanding what's captured in the query logs. I like this event-sourced pattern of tracking the run history of Queries.

https://github.com/chrisclark/django-sql-explorer/pull/562

chrisclark:
Addressed in https://github.com/chrisclark/django-sql-explorer/pull/562

chrisclark:
Happy to accept a separate PR for the script part that re-runs all queries and checks whether they will execute.

PR #555, opened 2023-10-19
Limit chart data by user specified number of rows

jamienoss:
@marksweb Hey, I was wondering if this is something you might be interested in?

We use this package and found it useful to graph only the result data as displayed and limited by the user specified row limit.

I added a setting for consideration for backward compatibility for existing users. If that's cool with you I'll incorporate this into the docs also.

I'll also add tests if you're cool with the changes in this PR.


chrisclark:
@jamienoss  I addressed this here:

https://github.com/explorerhq/sql-explorer/commit/5bb7c16130f01cbe622f99af2113d756b45800de

I took a slightly different approach, and simply respected the same row count restriction used in the preview pane.

Thanks for this good idea and apologies it took the better part of a year to get it integrated!

PR #556, opened 2023-11-02
build: Add readthedocs config

marksweb:
This will get RTD building again.

I seem to have more limited permissions than I did when I last released a version so I can't merge my own PRs or commit to master.


chrisclark:
Thanks!

PR #565, opened 2023-12-19
Front-end modernization

chrisclark:
The front-end tech stack has grown very stale. In addition to being on very old versions of Bootstrap and Codemirror, the front end also relies heavily on jQuery and other older technologies and techniques. 

The eventual ambition is to remove jQuery dependencies, and possibly build the interactive components of the app using...something. Possibly vue? HTMX? At this point there is no ambition to create a single-page-app, but some encapsulation of front-end components would surely benefit maintainability and testability.

So while this doesn't do all of that, it certainly modernizes a lot of stack and sets the project up for success along those fronts in the future. Major improvements in this refactor include:

- Moving to a Vite-based front-end build process for CSS and JS
- Move to SCSS
- Front-end deps are now managed via NPM
  - All client-side dependencies continue to be served locally
- Upgraded to Codemirror 6
- Upgraded to Bootstrap 5.3

Notable behavior changes include:

- Autocomplete support in the Codemirror editor, based on current connection's schema. Cool! It is *not* smart enough to switch SQL dialects based on the connection type, so all autocomplete is simply "generic sql" based. It also will not switch when the drop-down is changed; requires a page reload.
- 'Export to XLSX' functionality in the pivot table has been removed and instead uses a hand-rolled 'export to csv' functionality. This was to remove the dependency on xlsx.js, which was a rather large package providing minimal utility.
- Collapsing the SQL pane by default has been switched to a settings flag (EXPLORER_SHOW_SQL_BY_DEFAULT) and is set to True by default (e.g. the SQL pane will show for queries)

The dist JS and CSS files are not in the repo; they will be built as part of the release process (or can be built by any users locally by running `npm install` and `npm run build`).

To run the app locally, run `npm install`, then `npm run dev` (in a different terminal, or in the background, as it will block), then `python manage.py runserver`. These instructions are in the updated Readme as well.

TODOs:
- ~~QA of downloads, snapshots, charts, and more. Need a very thorough QA pass!~~
- ~~Pivot table JS is not working. Has to be added back.~~
- ~~XLSX export for pivot tables.~~
- ~~Key mappings and bindings for Codemirror~~
- ~~Codemirror auto-complete and other usability tweaks~~
- ~~Warn on unsaved changes~~
- ~~Prune old CSS and convert to SCSS~~
- ~~SQL pane toggle no longer works~~
- ~~Figure out how to deal with building the dist js and scss~~
- ~~Update Readme with good dev environment instructions~~
- ~~Whitespace (28 space characters) getting injected at the end of the Codemirror editor~~
- Update screenshots in readme


explorer/templates/explorer/query_list.html
```diff
@@ -4,44 +4,46 @@
...
-                        <td class="name">
-                            <a href="{% url 'query_detail' object.query_id %}">{{ object.query.title }}</a>
-                        </td>
-                        <td>{{ object.run_at|date:"SHORT_DATETIME_FORMAT" }}</td>
-                        <td class="text-center">
-                            <a href="{% url 'download_query' object.query_id %}">
-                                <i class="glyphicon glyphicon-download"></i>
-                            </a>
-                        </td>
+                        <th>{% trans "Query" %}</th>
```
marksweb, line 16: As we'll be transitioning to 5.0 support soon and [probably] dropping 3.2 as it's EOL, it might be worth updating these to `translate` as I'm not entirely sure when the `trans` alias will be dropped.

marksweb:
Exciting PR!! 🎉

marksweb:
Could you add some instructions on how to get this running?

I've used node 18 (it'd be good to put a `.nvmrc` file in for the version needed) and installed the node modules then ran `npm run build`.

But now I've got the build files in `public/static/dist`. Maybe I'm taking a look too soon, but I can't see the django static side of things that I think you had locally.

explorer/templates/explorer/query_list.html
chrisclark: Opened as separate issue:
https://github.com/chrisclark/django-sql-explorer/issues/567

marksweb (changes requested):

marksweb:
Running this again, UI is looking good now.

One thing that would be nice I think is a "home" link in the top right. Improves usability a little and it's a bit more obvious than clicking the app name. And if the heart on the favourites was red it might look a bit better & stand out.

Also, probably unrelated to these changes, but hitting "save & run" on queries is running them twice. My run counts are going from 0 to 2.

chrisclark:
Thanks for the feedback! I'll think more about the nav; there are some
other tweaks I'd like to make perhaps in another set of changes.

As for the 2x run thing -- I can't reproduce it. What might be happening is
you click on a query (viewing it - 1 run), then click save and run (another
run). And you are seeing the double runs that way.

I'll take one more good look through tomorrow and then...merge!

marksweb:
@chrisclark Oh of course - you open up a query and you see data, so there's run 1. 🤦 

Clearly too tired yesterday to think straight & figure that one out!!

chrisclark:
Alrighty - final set of changes here:

   - Squashed a bunch of commits and cleaned up messages. Still not great,
   but an improvement.
   - Added some visual indication of the home page being navigable, as you
   requested.
   - One other significant change: The schema is now injected into
   codemirror so it is able to autocomplete table and column names. Fancy!

Would you mind taking a final look, approving, and I will merge?

Thank you!

-Chris

explorer/static/config/vite.config.js
marksweb, line 1: Just going over the build steps, I don't think this should be within the application as it's not relevant to people installing the app via pip.

We should move the config dir to the root. I'm also thinking that it makes sense to have vite building to `explorer/static` so there's nothing more to that process of updating the static files for django.

explorer/static/js/src/codemirror-config.js
marksweb, line 1: I think these `src` files should live in a different directory.

Currently, if you ran `collectstatic` on a project you'd end up with these files being pulled into `static-collection`.

We'd be better off keeping the static files that aren't part of the final production files in something like `explorer/src/js` and `explorer/src/scss`. Then we can use those in the vite build to create the final files in `explorer/static/explorer/`. And that way the `static-collection` also isolates our files into a directory of the app name.

explorer/templates/explorer/fullscreen.html
```diff
@@ -1,26 +1,25 @@
 {% load i18n static %}
+{% load vite %}
```
marksweb, line 2: What I've mentioned about changes to the static files impacts on this template tag, but makes things a bit more like regular django.

Although maybe I don't know enough about vite, and maybe I'm just not doing things properly when I'm testing this. I've mostly been running the vite build, then copying the files into the static directory and then the app into a project like it would if it'd come from pypi.

package.json
```diff
@@ -0,0 +1,45 @@
...
+    "stylelint-config-recess-order": "^4.0.0",
+    "stylelint-config-standard-scss": "^11.0.0",
+    "stylelint-scss": "^5.0.0",
+    "vite": "^5.0.1"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/chrisclark/django-sql-explorer.git"
+  },
+  "author": "Brent O'Connor",
```
marksweb, line 29: I'd update this

package.json
```diff
@@ -0,0 +1,45 @@
...
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/epicserve/django-base-site/issues"
```
marksweb, line 32:
```suggestion
    "url": "https://github.com/chrisclark/django-sql-explorer/issues"
```

package.json
```diff
@@ -0,0 +1,45 @@
...
+  },
+  "homepage": "https://github.com/epicserve/django-base-site#readme",
```
marksweb, line 34:
```suggestion
  "homepage": "https://github.com/chrisclark/django-sql-explorer#readme",
```

pypi-release-checklist.md
```diff
@@ -3,7 +3,8 @@
 - [x] Make sure any new files are included in MANIFEST.in
 - [x] Update version number in `explorer/__init__.py`
 - [x] Update any package dependencies in `setup.py`
-- [x] Commit the changes: 
+- [x] Run `npm install` and `npm run build`
```
marksweb, line 6: Need `nvm use` (and `nvm install` before that) before you do the `npm install`.

chrisclark:
These are awesome changes. I'm on it!

Thanks to your hints above, I also figured out why it's not really working
for you locally. Totally my fault. Two issues and two fixes:

1. *Having to build the files.* The vite template tag looks for
`settings.VITE_DEV_MODE` to be truthy. When that is the case, it will do
some magic to load the files from the Vite dev server (which will have to
be running). Then, live reload works, source inspection works, etc. But if
VITE_DEV_MODE is not set/false, then the template tag will look to load the
generated main-<gibberish>.js built file. Because I didn't document that,
of course you didn't do it, and thus had to resort to building the release
versions of the files. Set that VITE_DEV_MODE setting to true, and `npm run
dev`, and your front-end dev experience will be way more awesome :)

2. *Having to copy the files.* Again my fault for putting them in a
weird /public/ directory. If you had something like:

STATICFILES_DIRS = STATICFILES_DIRS.append('./explorer/public/static')

Then the built files would get collected. But I'll normalize the directory
structure as you suggest so it works 'out of the box'.

-Chris

marksweb:
I did try to run the project with vite dev mode enabled in django. I may have been doing it wrong of course.

I copied the app dir (`explorer`) into the project where dev mode was enabled. Then I had downloaded the zip of the project and had it in my downloads folder where I had the dev server running. 

And I saw the test project setup to look for static files there, but yeah it'd be better to avoid that & keep things more like regular django.

chrisclark:
All of these changes have been made! Woohoo!

I'm going to dogfood some of the instructions/docs/readme and make sure
things are really working --- but perhaps some of your issues running it
locally have been resolved as well?? If not, hopefully I'll be able to
reproduce those issues soon as I do the aforementioned testing.

chrisclark:
Had a nice little debugging session just now! Still had some path issues
which have now been resolved. I'm feeling confident that things are working
in a sane fashion.

I downloaded the branch as a zip into a new folder, extracted it, and ran:

pip install -r requirements/base.txt
pip install django
python manage.py migrate
python manage.py createsuperuser
python manage.py runserver

(and in a separate terminal)

nvm use
npm install
npm run dev

And voila - works like a charm locally. I can also stop the vite server, do
npm run build, flip the VITE_DEV_MODE flag, and everything loads.

I also ran a full test following the instructions (updated!) in
pypi-release-checklist.md (which involves creating an sdist bundle, a new
django project, installing it, etc) and all works well there too.

It was a pain in the ass getting all of the relative directories working!
Primarily related to how to read the manifest file in a reasonably
high-confidence way.

I'm feeling quite good about the state of things. If you still have any
issues, I'd really love to know as I do believe it should "just work".
Thank you so much!

PR #566, opened 2023-12-20
ci: Add testing for django 5.0 and django main

chrisclark:


marksweb:
@chrisclark Here we go. Tests all passing, github action now running the coverage reporting and tests via tox.

Testing now done against;

- Python 3.10, 3.11, 3.12
- Django 4.2, 5.0, main

All ran against the base dependencies and then the optional ones.

chrisclark:
Terrific!

PR #577, opened 2024-01-31
4.0.0.beta1

chrisclark:
Fix a few outstanding front-end bugs, and add this helpful message if someone pulls the project and runs it without running Vite.

image


chrisclark:
@marksweb I went as far as writing a github action that would build and commit the front-end in master -- but I just didn't like it. I hate the idea of having that built file in the repo. I think this is a somewhat silly, but legitimately helpful, move in the right direction.

I'm also going to ready the 4.0 release in this branch.

chrisclark:
@marksweb I also removed the pyproject.toml file because it's quite confusing to have it when the build process is still controlled by setup.py. It appears it was being used just for ruff config (and in fact had some conflicting coverage config in it) so I moved it to a dedicated ruff.toml file.

explorer/__init__.py
```diff
@@ -1,7 +1,7 @@
 __version_info__ = {
-    "major": 3,
-    "minor": 2,
-    "patch": 1,
+    "major": 4,
+    "minor": 0,
+    "patch": 0,
     "releaselevel": "final",
```
marksweb, line 5: Might be worth doing a beta release first, just because the changes are so big.

explorer/templates/explorer/base.html
```diff
@@ -20,6 +20,27 @@
 </head>
 
 <body>
+{% if vite_dev_mode %}
+    <div class="vite-not-running-canary" style="text-align:center;">
```
marksweb, line 24: Is the idea here that is vite is running, the CSS for this class is available and therefore this gets hidden?

.github/workflows/publish-pypi.yml
```diff
@@ -21,6 +21,14 @@ jobs:
       with:
         python-version: '3.12'
 
+    - uses: actions/setup-node@v4
```
marksweb, line 24: Yes, this is the right idea. The built static files are only needed by people who are installing the app via pypi, so this will create the files they need and we just need to make sure that they end up in the built package and will get copied by `collectstatic` by projects.

explorer/__init__.py
chrisclark: Yeah, probably a good idea. I will do that. And I'll get my company to adopt it and test it :)

explorer/templates/explorer/base.html
chrisclark: Yes exactly! It works well and is reasonably helpful I think. See the screenshot.

marksweb (approved):
Ok I think this looks good for a beta release 👍

marksweb:
Ok, so merging this will put the beta on here; https://test.pypi.org/project/django-sql-explorer/

Then creating a release [here](https://github.com/chrisclark/django-sql-explorer/releases) for the beta will put a pre-release version on pypi; https://pypi.org/project/django-sql-explorer/ (there's a checkbox for pre-release on the release creation)

chrisclark:
@marksweb Done -- but yikes! It appears to have pushed a non-beta 4.0 version to https://pypi.org/project/django-sql-explorer/

It's not clear what I did wrong...the version is a beta in __init__, I hit the 'pre release' checkbox. Any ideas?

marksweb:
@chrisclark just on my phone, out of the house right now so can't dig into it, but the release you created is set against a commit that's a v4 beta and `setup.py` imports it's version from `explorer` so there's nothing obvious.

chrisclark:
Ok. I'll poke around and see if I can get it working before I push beta2 later today or tomorrow.

PR #578, opened 2024-02-01
4.0 Beta 2

chrisclark:
- Removes jquery-ui dependency (was easy to do)
- Uses dynamic imports so the JS is now chunked
- Fixes a subtle versioning info that prevents PyPi from flagging beta releases
- Adds back support for Django 3.2 (no code changes required - simply added back to tox tests and setup.py)
- Fixes a few styling bugs


marksweb:
3.2 hits end of life in April so not sure it's worth too much effort to put these back in

chrisclark:
For sure. But if supporting it is "free" (e.g. no code changes required) then I'd like to do it. But I'm not going to bend-over backwards.

I think one of the principles of this project is that "it just works". It's killing me that the front-end build has sort of broken that, but in general that's my thinking. A bunch of folks still run 3.2 (especially internally) and so if it's easy, I'd just as soon support it (until it's not easy).

marksweb:
Yes absolutely agree with that 👍

marksweb:
@chrisclark It looks like this release is ok? I can see on pypi that it's 4.0b2 and looks like the beta 1 ended up as 4.0 on pypi. 

So that `get_version` is just a bug waiting to happen when you want to push a release that isn't final because you have to disable `short` and you were only to know that after publishing what you thought was a beta.

That check should probably become something like this to ensure that non "final" releases work;

```
    if __version_info__["releaselevel"] != "final":
        vers.append(
            "%s%i" % (
                __version_info__["releaselevel"][0],
                __version_info__["serial"])
        )
```

chrisclark:
Ah! Check out the beta2 PR. I figured out why beta1 wasn’t published correctly on pypi and fixed it! You are on the right track, but the issue was really in setup.py, not in __init__. The ‘short=True’ parameter was just being passed into the wrong function in __init__.

PR #582, opened 2024-02-05
4.0b3

chrisclark:
- When testing beta 3, discovered that static URLs other than /static/ are not working. This addresses the issue.
- It also greatly simplifies the inclusion of the built front-end by not versioning the files, and therefore not requiring the manifest.


explorer/templatetags/vite.py
```diff
@@ -1,61 +1,40 @@
...
 from django import template
 from django.conf import settings
 from django.utils.safestring import mark_safe
 from explorer import app_settings
 
-
 register = template.Library()
 
-VITE_OUTPUT_DIR = "/static/explorer/"
+VITE_OUTPUT_DIR = f"{settings.STATIC_URL}explorer/"
```
marksweb, line 10: I've never worked with a project that didn't use `/static` so I didn't even think about that 🤦

PR #593, opened 2024-02-20
Checking if the settings.AUTH_USER_MODEL has email field or not!

abedimhosein:
If `settings.AUTH_USER_MODEL` is a custom user model that does not have an email field, the `created_by_user_email` function throws an AttributeError exception.

### DIFF
--- a/explorer/forms.py
+++ b/explorer/forms.py
@@ -29,7 +29,6 @@ class SqlField(CharField):
 
 
 class QueryForm(ModelForm):
-
     sql = SqlField()
     snapshot = BooleanField(widget=CheckboxInput, required=False)
     connection = CharField(widget=Select, required=False)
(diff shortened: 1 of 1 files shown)


explorer/forms.py
```diff
@@ -49,8 +48,14 @@ def clean(self):
 
     @property
     def created_by_user_email(self):
-        return self.instance.created_by_user.email if \
-            self.instance.created_by_user else "--"
+        user = self.instance.created_by_user
```
marksweb, line 51: Might be neater without assigning the user first;

```
            try:
                return self.instance.created_by_user.email
            except AttributeError:
                return "--"
```

You could also set a property on your custom user called `email` that returned some way of identifying your users. That would avoid attribute errors with third party apps.

marksweb:
It'd be good to see some test coverage of this one.

explorer/forms.py
marksweb: Just thinking about this again, perhaps it makes more sense, if a user doesn't have an email address, just return the string representation of a user.

```
try:
    return self.instance.created_by_user.email
except AttributeError:
    return str(self.instance.created_by_user)
```

Then this gives projects that don't have an `email` field the chance to add an `email` property to do something custom, or just display the default representation of a user object.

abedimhosein: Yes, your solution is better. By the way, I'm just wondering why you are using the email and not using the string representation of the user in the first place?

chrisclark:
Closing in favor of:
https://github.com/chrisclark/django-sql-explorer/pull/606

Issue #595, opened 2024-02-26
Collect anonymous profiling stats


chrisclark:
An early heads up that this is coming! To figure out where to go from here, I'd like some profiling data about how the application is being used. In an upcoming release, we will collect some basic anonymous usage statistics.

chrisclark:
Working on it:
https://github.com/chrisclark/django-sql-explorer/pull/596

PR #596, opened 2024-02-29
Anonymous usage data collection (issue #595)

chrisclark:
Gathers a variety of usage metrics (largely at application start-up) and sends to a collection endpoint.

Some of the more interesting bits of this are:
- Generating a stable, but unique, ID per 'install'. See explorer.tracker._instance_identifier()
- Running stat collection at app start time. Strangely, Django doesn't have a good way to do this. See explorer.apps.track_summary_stats()
- The HTTP requests that are sent to the collection endpoint are 'fire and forget' via a thread.
- New app setting - EXPLORER_ENABLE_ANONYMOUS_STATS - that can be set to False to disable data collection

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ explorer/static/
 # Sphinx documentation
 docs/_build/
 .env
+tst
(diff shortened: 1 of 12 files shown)


chrisclark:
This PR is stuck in a weird state. Going to close and reopen.

PR #597, opened 2024-03-01
Anonymous telemetry

chrisclark:
Gathers a variety of usage metrics (largely at application start-up) and sends to a collection endpoint.

Some of the more interesting bits of this are:
- Generating a stable, but unique, ID per 'install'. See explorer.tracker._instance_identifier()
- Running stat collection at app start time. Strangely, Django doesn't have a good way to do this. See explorer.apps.track_summary_stats()
- The HTTP requests that are sent to the collection endpoint are 'fire and forget' via a thread.
- New app setting - EXPLORER_ENABLE_ANONYMOUS_STATS - that can be set to False to disable data collection

Other changes:
- Improved documentation (dev instructions, testing instructions, celery setup)
- Factored out a new tests.settings file so that test_project/settings.py is not overloaded to both serve as a good demo of the app AND supply all test values (updated tox accordingly)

### DIFF
--- a/explorer/apps.py
+++ b/explorer/apps.py
@@ -1,6 +1,7 @@
 from django.apps import AppConfig
 from django.core.exceptions import ImproperlyConfigured
 from django.db import connections as djcs
+from django.db.utils import OperationalError
 from django.utils.translation import gettext_lazy as _
 
 
@@ -14,6 +15,7 @@ class ExplorerAppConfig(AppConfig):
         from explorer.schema import build_async_schemas
         _validate_connections()
         build_async_schemas()
+        track_summary_stats()
 
 
 def _get_default():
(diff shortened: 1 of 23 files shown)


explorer/apps.py
```diff
@@ -51,12 +51,14 @@ def track_summary_stats():
...
     from explorer.models import Query
 
-    # Test to see if migrations have been applied.
-    # If not, then we won't gather start-up stats (no need -- there's nothing interesting to report on)
-    # We can't use a post_migrate hook because in many cases the app will be starting
-    # and there won't be any migrations running! So the signal would never get called.
-    # Django doesn't actually have a way of running code on application initialization,
-    # so we are kinda faking it by doing this.
+    # Django doesn't actually have a way of running code on application initialization, so we have come up with this.
+    # The app.ready() method (the call site for this function) is invoked *before* any migrations are run. So if were
```
marksweb, line 55: It sounds like maybe you want to have a signal reciever that runs on the `post_migrate` signal.

https://docs.djangoproject.com/en/5.0/ref/signals/#post-migrate

I did something similar for django-cms where it uses `pre_migrate` to prevent people from accidentally migrating their database;
https://github.com/django-cms/django-cms/blob/develop-4/cms/signals/__init__.py#L35

docs/requirements.txt
```diff
@@ -3,3 +3,4 @@ furo
 Sphinx>4
 sphinx-copybutton
 sphinxext-opengraph
+requests>=2.2
```
marksweb, line 6: Don't think this is needed in here as these are requirements specific to building the docs.

marksweb (changes requested):

explorer/apps.py
chrisclark: No -- keep reading the comment. I tried that, but if there are no actual migrations to run, then post_migrate never gets called. The intent of this is to grab summary stats on every application start-up -- many of which will not have net-new migrations and therefore never fire the signal.

PR #603, opened 2024-04-18
Refactored reqs

Anton-Shutik:
Let's split the reqs into files and keep it in one place


tox.ini
```diff
@@ -2,20 +2,20 @@
...
 allowlist_externals = coverage
 deps =
-    base-reqs: -r requirements/base.txt
+    base-reqs: -r requirements/dev.txt
     dj32: django>=3.2,<4.0
     dj42: django>=4.2,<5.0
     dj50: django>=5.0,<5.1
     djmain: https://github.com/django/django/archive/main.tar.gz
-    optional: -r requirements/optional.txt
+    dev: -r requirements/dev.txt
```
Anton-Shutik, line 18: What the idea of testing base reqs and optional ? Seems like base reqs should not include test ones, and thus it makes impossible to run tests on base-reqs.

cc @chrisclark

chrisclark: I agree with that.

chrisclark (approved):
This is a really nice improvement. Thank you!

Issue #615, opened 2024-04-30
Has there any migration to apply?


jefer94:
Python: **3.12.3**
Lib: **4.2**

My tests broke when I tried to delete an `auth.User`, here is the traceback:

```py
________________________________ AuthenticateTestSuite.test__post__with_token__right_password__passing_callback _________________________________
../../../.local/share/virtualenvs/apiv2-bqPrB3_w/lib/python3.12/site-packages/django/db/backends/utils.py:105: in _execute
    return self.cursor.execute(sql, params)
../../../.local/share/virtualenvs/apiv2-bqPrB3_w/lib/python3.12/site-packages/newrelic/hooks/database_dbapi2.py:37: in execute
    return self.__wrapped__.execute(sql, parameters,
../../../.local/share/virtualenvs/apiv2-bqPrB3_w/lib/python3.12/site-packages/django/db/backends/sqlite3/base.py:329: in execute
    return super().execute(query, params)
E   sqlite3.OperationalError: no such table: explorer_promptlog

The above exception was the direct cause of the following exception:
breathecode/authenticate/tests/urls/tests_password_token.py:459: in test__post__with_token__right_password__passing_callback
    self.bc.database.delete('auth.User')
breathecode/tests/mixins/breathecode_mixin/database.py:561: in delete
    return model.objects.filter(**lookups).delete()
../../../.local/share/virtualenvs/apiv2-bqPrB3_w/lib/python3.12/site-packages/django/db/models/query.py:1187: in delete
    collector.collect(del_query)
../../../.local/share/virtualenvs/apiv2-bqPrB3_w/lib/python3.12/site-packages/django/db/models/deletion.py:348: in collect
    if getattr(on_delete, "lazy_sub_objs", False) or sub_objs:
../../../.local/share/virtualenvs/apiv2-bqPrB3_w/lib/python3.12/site-packages/django/db/models/query.py:414: in __bool__
    self._fetch_all()
../../../.local/share/virtualenvs/apiv2-bqPrB3_w/lib/python3.12/site-packages/django/db/models/query.py:1928: in _fetch_all
    self._result_cache = list(self._iterable_class(self))
../../../.local/share/virtualenvs/apiv2-bqPrB3_w/lib/python3.12/site-packages/django/db/models/query.py:91: in __iter__
    results = compiler.execute_sql(
[...]

PR #618, opened 2024-05-10
export assistant models

MartingaleCoda:
Fixes issue where `python manage.py makemigrations` expects PromptLog model to be deleted, as the explorer app doesn't include it in its `models.py` file.

Example of the issue when moving from 4.0.2 to 4.2.0:
```
python3 manage.py makemigrations --check
Migrations for 'explorer':
  /opt/venv/lib/python3.9/site-packages/explorer/migrations/0016_delete_promptlog.py
    - Delete model PromptLog
```

Issue [615](https://github.com/chrisclark/django-sql-explorer/issues/615) may be related.

### DIFF
--- a/explorer/assistant/__init__.py
+++ b/explorer/assistant/__init__.py
@@ -1,0 +1,1 @@
+from .models import PromptLog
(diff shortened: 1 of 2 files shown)


chrisclark:
Ahh, that makes sense. Thanks so much for clarifying. Yeah I'll bet that is source of the issue in #615. I'll have to figure out why I didn't catch it in testing...

Will get this into a point release.

chrisclark:
Merged. Thanks again!

PR #619, opened 2024-05-15
Fix CSRF when CSRF_USE_SESSIONS is enabled

cpmsmith:
The front-end currently checks for the `input[name="csrfmiddlewaretoken"]` element if `CSRF_COOKIE_HTTPONLY` is set, but according to the [Django docs], it should do the same thing if `CSRF_USE_SESSIONS` is set, as it will also mean the token is not available in a cookie.

[Django docs]: https://docs.djangoproject.com/en/5.0/howto/csrf/#acquiring-the-token-if-csrf-use-sessions-or-csrf-cookie-httponly-is-true

### DIFF
--- a/explorer/src/js/csrf.js
+++ b/explorer/src/js/csrf.js
@@ -1,10 +1,10 @@
 import cookie from "cookiejs";
 
 const csrfCookieName = document.getElementById('csrfCookieName').value;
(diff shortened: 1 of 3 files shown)


chrisclark:
Thanks! I will get it reviewed and merged into the next release.

PR #620, opened 2024-05-16
Ug remove future imports from explorer

Ujjwal-Squadstack:


Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+
+# Set the working directory in the container
+WORKDIR /app
+
+# Copy the requirements file into the container at /app
+COPY requirements.txt /app/
+COPY optional-requirements.txt /app/
+
+# Install any needed packages specified in requirements.txt
+RUN pip install Django==1.11.17
```
marksweb, line 16: 🧐

marksweb (commented):
Some context or conversation as to the source of this PR would be useful because it's doing a lot more than removing future imports.

chrisclark:
> Some context or conversation as to the source of this PR would be useful because it's doing a lot more than removing future imports.

@marksweb i think this got opened by mistake, hah. It is a very long running (and fairly active!) fork of a very old version of the project.

PR #626, opened 2024-06-06
Light branding, and bugixes to user uploads

chrisclark:


explorer/schema.py
```diff
@@ -53,7 +53,10 @@ def schema_json_info(connection_alias):
     ret = cache.get(key)
     if ret:
         return ret
-    si = schema_info(connection_alias) or []
+    try:
+        si = schema_info(connection_alias) or []
+    except InvalidExplorerConnectionException as e:
```
marksweb, line 58: No need to `as` this if it's not then going to raise an exception. If it was you could then do `raise Exception from e`

marksweb: 😆 I spotted an email come in that mentioned branding so I was curious & just scrolling through. The ruff alerts made it easy to see what it was complaining about!

What branch would be good for me to run?

I was looking recently at building a django app with explorer for work that had it's own sqlite database to handle the project tables, but then an extra connection that used google cloud SQL proxy to connect to our remote databases for running queries.

Currently we create that proxy connection from the CLI and then I use a tool like [TablePlus](https://tableplus.com/) to query the DB.

chrisclark: This branch, for sure. I ended up finding a bunch of bugs that I'll fix in another beta release.

Make sure these are set up in settings:

EXPLORER_DB_CONNECTIONS_ENABLED = True
EXPLORER_S3_BUCKET = os.environ.get("EXPLORER_S3_BUCKET")
EXPLORER_S3_ACCESS_KEY = os.environ.get("EXPLORER_S3_ACCESS_KEY")
EXPLORER_S3_SECRET_KEY = os.environ.get("EXPLORER_S3_SECRET_KEY")

And you will see the new "connections" nav item. You can add a new connection and enter credentials, or drag and drop a CSV or sqlite file (bottom of the new connection page). Let me know what you think! And of course if you run into any issues.

The 'branding' is like "Let's make this look a little less shitty and like default bootstrap".

chrisclark: image

Issue #633, opened 2024-07-02
Creates migration in dependencies directory


spapas:
Hello, after installing the version 5.0.1 I get the following warning when running `migrate`:

```bash
  Your models in app(s): 'explorer' have changes that are not yet reflected in a migration, and so won't be applied.
  Run 'manage.py makemigrations' to make new migrations, and then re-run 'manage.py migrate' to apply them.
```

When I run makemigrations it creates me a migration file in the dependency directory (i.e C:\progr\py3\ehcg\venv\Lib\site-packages\explorer\migrations). The migration file is `0019_alter_databaseconnection_engine.py` with the following contents :

```python

# Generated by Django 5.0.6 on 2024-07-02 06:55

from django.db import migrations, models


class Migration(migrations.Migration):

    dependencies = [
        ("explorer", "0018_alter_databaseconnection_host_and_more"),
    ]

    operations = [
        migrations.AlterField(
            model_name="databaseconnection",
            name="engine",
            field=models.CharField(
                choices=[
                    ("django.db.backends.sqlite3", "SQLite3"),
                    ("django.db.backends.postgresql", "PostgreSQL"),
                    ("django.db.backends.mysql", "MySQL"),
                    ("django.db.backends.oracle", "Oracle"),
                    ("django.db.backends.mysql", "MariaDB"),
                    ("django_cockroachdb", "CockroachDB"),
                    ("django.db.backends.sqlserver", "SQL Server (mssql-django)"),
                ],
                max_length=255,
            ),
        ),
    ]

```

This is very problematic because when a new migration will be added to django-sql-explorer it will also get the number 19 and there will be a conflict. Can you please advise on how this situation should be handled ? 

Thank you !

jefer94:
I'm having the same error https://github.com/breatheco-de/apiv2/actions/runs/9757660712/job/26930572478?pr=1404

chrisclark:
Yep - I am addressing right now: #634

chrisclark:
Will be released in the next ~30 mins as 5.0.2 on pypi. Apologies for the oversight. I will also look to add a CI check to prevent this in the future.

jefer94:
I have this

# .github/workflows/checks.yml
/checks.yml
```yml

  migrations:
    needs: cache
    runs-on: ubuntu-latest

    steps:
      - name: Checkout code
        uses: actions/checkout@v3

      - name: Set up Python
        uses: actions/setup-python@v4
        with:
[...]

PR #635, opened 2024-07-03
CI to detect out of sync migrations

chrisclark:
Making sure #633 doesn't happen again.


.github/workflows/test.yml
```diff
@@ -53,6 +53,9 @@ jobs:
           name: coverage-data-${{ matrix.python-version }}
           path: '${{ github.workspace }}/.coverage*'
 
+      - name: Check migrations
+        run: python manage.py makemigrations --check
```
marksweb, line 57: You can add this to the test suite as well.

https://adamj.eu/tech/2024/06/23/django-test-pending-migrations/

marksweb: Actually it's already part of the pre-commit hooks so in theory you shouldn't be able to commit code that needs migrations and doesn't include them.

jefer94: Pre-commit should be bypassed with a param on the console if someone didn't install Pre-commit

chrisclark: I'd like to address this particular issue outside of precommit. Interesting idea to do it in tests directly...I might go that direction seeing as it's a bit janky to put it in the pypi-test as I've done here (just did it to save time during CI to piggyback on the environment)

chrisclark:
This turned into a pretty interesting PR. It turns out that tox was installing all of the dev dependencies for all of the various configurations -- which is why (for example) I had an import issue in the 5.0 release (a dev dependency turned out to be required, even for a base install). I've reconfigured the tox setup to properly test with JUST the base deps, and then also with 'everything'. This gets better coverage on the 'everything' version (hits more test cases) and runs faster (because it's actually installing only the base deps for half of the test runs). Cool! Also required skipping certain tests for the base tests -- and I actually caught another import problem with the base install!

.github/workflows/test.yml
marksweb: > Pre-commit should be bypassed with a param on the console, and if someone didn't install Pre-commit

Yes we need pre-commit either running manually as a workflow run, or integrating the bot from pre-commit.ci

I think we've already got PRs from people who don't have pre-commit installed.

PR #644, opened 2024-07-09
Dockerfile

chrisclark:
This adds a dockerfile to allow the project to run with a single command. It will replace start.sh.


marksweb:
Probably worth having a compose file as well? Unless I'm missing something here.

chrisclark:
@marksweb yes - i agree. Do you think by default this should include e.g. tasks (redis + celery), user uploads, etc? Should it be maximal or minimal?

marksweb:
@chrisclark If it's for giving people an example, then the more the better I suppose. So could have a full setup, or something more complex that offers the different options from a basic setup through to all the options.

chrisclark:
@marksweb i am getting frustrated with the docker-compose file. It has some issue when running the vite dev server, that is not present in the dockerfile. It's very annoying. Basically running this:

`docker run -p 8000:8000 -p 5173:5173 django-sql-explorer`

Works fine, but running:

`docker-compose up`

Leads to this error:

```
web-1  | Django server started
web-1  | Starting Vite with APP_VERSION=5.0.2
web-1  | /app/node_modules/rollup/dist/native.js:59
web-1  | 		throw new Error(
web-1  | 		      ^
web-1  |
web-1  | Error: Cannot find module @rollup/rollup-linux-arm64-gnu. npm has a bug related to optional dependencies (https://github.com/npm/cli/issues/4828). Please try `npm i` again after removing both package-lock.json and node_modules directory.
web-1  |     at requireWithFriendlyError (/app/node_modules/rollup/dist/native.js:59:9)
web-1  |     at Object.<anonymous> (/app/node_modules/rollup/dist/native.js:68:76)
web-1  |     ... 3 lines matching cause stack trace ...
web-1  |     at Module._load (node:internal/modules/cjs/loader:1019:12)
web-1  |     at ModuleWrap.<anonymous> (node:internal/modules/esm/translators:203:29)
web-1  |     at ModuleJob.run (node:internal/modules/esm/module_job:195:25)
web-1  |     at async ModuleLoader.import (node:internal/modules/esm/loader:337:24) {
web-1  |   [cause]: Error: Cannot find module '@rollup/rollup-linux-arm64-gnu'
```

I've messed around with it for a while now and can't get it to work and am inclined to just give up for now, and leave this open if anyone wants to pick it up.

chrisclark:
Alright, I now know the issue -- basically the docker-compose mounts the local filesystem into the /app volume, which steps on the node_modules directory *within* the container, with the host filesystem's node_modules directory. In theory this could be resolved by adding a named mounted volume for node_modules -- but goddamit I can't get it to work!

Eg. this approach should really work:
https://github.com/vitejs/vite/discussions/15532#discussioncomment-8038794

But I just can't get it to. I get an error about 'the volume is busy' when trying to clean and re-install the npm modules.

This is super frustrating because obviously this is a totally simple docker setup. Frustrating.

marksweb:
@chrisclark Interesting problem - and a niche one to figure out.

Perhaps a `.dockerignore` file might help to exclude `node_modules` - though I don't know if this might be conflicting/problematic for the volume mounting of compose.

chrisclark:
With the clarity of it not being midnight, I got this working! I added a volume to the docker-compose specifically for node_modules, removed the npm clean and install from the entrypoint script (so if npm dependencies change, the docker container will have to be rebuilt), and added node_modules to dockerignore. Phew! I swear I tried this last night and it didn't work -- but now it does! 

docker-compose build
docker-compose up

Voila!

I'll extend this a bit more when I have some time.

Thanks for the encouragement @marksweb

marksweb:
Excellent news!

And bare in mind that `docker-compose` may stop working soon - latest mac install of docker has removed it. You'll have to get used to doing `docker compose` as it's been part of docker core for a while.

PR #649, opened 2024-07-22
Use a more reliable source of the static files URL

illia-v:
I have a case when SQL explorer pages don't render correctly because paths of all static files are wrong.

django-storages is used and [`AWS_LOCATION`](https://django-storages.readthedocs.io/en/latest/backends/amazon-S3.html#:~:text=location%20or%20AWS_LOCATION) is set, so actual paths are prefixed.
The SQL explorer uses `settings.STATIC_URL` to construct URLs of static files, but `settings.STATIC_URL` doesn't contain the prefix in the described configuration.

There may be other cases when storage classes change URLs, let's build the URLs using a configured Django storage.

### DIFF
--- a/explorer/templatetags/vite.py
+++ b/explorer/templatetags/vite.py
@@ -2,12 +2,13 @@ import os
 
 from django import template
 from django.conf import settings
+from django.contrib.staticfiles.storage import staticfiles_storage
 from django.utils.safestring import mark_safe
 from explorer import app_settings, get_version
 
 register = template.Library()
 
-VITE_OUTPUT_DIR = f"{settings.STATIC_URL}explorer/"
+VITE_OUTPUT_DIR = staticfiles_storage.url("explorer/")
 VITE_DEV_DIR = "explorer/src/"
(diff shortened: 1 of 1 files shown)


chrisclark:
Love this! Thank you.

I have struggled to get all of the various django static stuff to be really solid, and I think this helps. I will review shortly, merge, and get a beta release out by the end of this week (along with a bunch of other improvements). I'll get it tested in a more complex environment next week and hopefully have a final release done by Aug 2.

illia-v:
@chrisclark thank you for reviewing this and providing the expected release date!

chrisclark:
@illia-v FYI 5.1b1 is on pypi:
https://pypi.org/project/django-sql-explorer/5.1b1/

Will try to cut a final release sometime in the coming week.

chrisclark:
I have tested this on demo.sqlexplorer.io, with similar django-storages settings:

https://github.com/explorerhq/demo-site/blob/master/explorer_demo/settings.py

And it seems to be working well. I will cut a release tomorrow after one more round of testing in another environment.

PR #651, opened 2024-07-24
Advanced uploads

chrisclark:
- Ability to append an upload to a previously uploaded file/sqlite DB as a new table
- Good cache busting and detection of file changes on uploads
- Separate the upload UI from the 'add connection' UI, as they are materially different
- Fix a small bug with bar chart generation, when values are null
- Ability to refresh a connection's schema and data (if it's an upload) from the connections list view

TODO
- ~~Increase test coverage on append (e.g. does the table get overwritten when it already exists? what happens when the append ID val in the POST request is invalid? Are there race conditions that could happen?)~~
- ~~Create some kind of check that invalidates the downloaded SQLite files. If a server has a copy of the original SQLite uploaded file, but then someone appends a table to it (on a different node/server), the first server will never go get the updated version from s3.~~
- ~~Similarly, we might need to invalidate schema data when databases have a file appended.~~
- ~~Documentation~~


docs/features.rst
```diff
@@ -169,7 +169,49 @@ Multiple Connections
   multi-connection setup.
 - SQL Explorer also supports user-provided connections in the form
   of standard database connection details, or uploading CSV, JSON or SQLite
-  files. See the 'User uploads' section of :doc:`settings`.
+  files.
+
+File Uploads
```
chrisclark, line 174: @marksweb If you have a moment, I'd love for you to give this a read and let me know what you think. Not asking for a full code review, just looking over the docs and make sure the functionality is clear to you. If there are outstanding questions, I can address them in the docs, or perhaps just by reading the docs you will think of some edge case or test I haven't thought of! I would welcome a brain-dump from you on this functionality.

PR #660, opened 2024-08-13
All connections are db-backed

chrisclark:
A simplification of how "connections" work. Instead of a weird blend of DatabaseConnection models and underlying Django models, instead insist that all Queries and QueryLogs are foreign-keyed to an actual DatabaseConnection model.

There is a migration that will create new DatabaseConnection models that wrap the underlying Django connection, to preserve compatibility. The config of those connections is still read from settings.py, making this a 'safe' migration, regardless of how the underlying connections were set up.


explorer/assistant/utils.py
```diff
@@ -127,9 +126,8 @@ def fits_in_window(string: str) -> bool:
 
 def build_prompt(request_data, included_tables):
     user_prompt = ""
-
-    db_vendor = get_valid_connection(request_data.get("connection")).vendor
-    user_prompt += f"## Database Vendor / SQL Flavor is {db_vendor}\n\n"
+    djc = DatabaseConnection.objects.get(id=request_data.get("connection")).as_django_connection()
```
marksweb, line 129: Is this guaranteed to return an object? If not, it should either `get_or_create` or do something defensive with `try/except`

PR #696, opened 2025-01-31
Gracefully handle NotSupportedError during schema introspection

gavinhodge:
In certain circumstances, a django.db.utils.NotSupportedError can be thrown during schema introspection. If this happens, the exception is unhandled and Django returns a 500 response instead of displaying query results.

To reproduce requires:

- Postgresql running in recovery mode (typically a hot standby server). This can be [simulated](https://dba.stackexchange.com/questions/80620/how-to-simulate-postgresql-recovery-mode) without a hot standby server
- An unlogged table on the primary (`ALTER TABLE table_name SET UNLOGGED`)
- Attempt to execute a query (e.g. `SELECT 1`) against the standby server and receive a 500 error page

Exception detail:
```
NotSupportedError
cannot access temporary or unlogged relations during recovery
```

The issue can be worked around by specifying the affected tables in the `EXPLORER_SCHEMA_EXCLUDE_TABLE_PREFIXES` setting, but this change makes a workaround unnecessary.

### DIFF
--- a/explorer/schema.py
+++ b/explorer/schema.py
@@ -1,5 +1,5 @@
 from django.core.cache import cache
-from django.db import ProgrammingError
+from django.db import NotSupportedError, ProgrammingError
 
 from explorer.app_settings import (
     EXPLORER_SCHEMA_EXCLUDE_TABLE_PREFIXES,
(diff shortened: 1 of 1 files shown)


gavinhodge:
Hi maintainers, does this need any action from my end before it can be reviewed or merged?

PR #703, opened 2025-03-13
Remove the x-requested-with check from EmailCsvQueryView

### LINKED ISSUE #702: Queueing a report to be sent via email doesn't work
Currently when you try and have a report emailed, the request from the send modal to the server fails:

Image

The expectation that isn't being met is that the front-end isn't passing an "X-Requested-With" header, which is causing this logic to fail:

https://github.com/explorerhq/sql-explorer/blob/910cb710ab4015d3ad32e4c63b9263e7381bd19f/explorer/views/email.py#L12-L20

I believe this was caused by this commit, which switched from using jQuery, which emits an `X-Requested-With` header, to fetch, which does not:

https://github.com/explorerhq/sql-explorer/commit/c4278c23dda8b430dfdd9b452d9d938510f450ce#diff-c2106f38c65afec174958944fc7f073b461b2ae16929467de22e49e76e6055e4L56-R63

jarcoal:
I don't believe this check offers any security benefits as the view requires a CSRF token anyway.

Resolves #702

### DIFF
--- a/explorer/tests/test_views.py
+++ b/explorer/tests/test_views.py
@@ -836,28 +836,17 @@ class TestEmailQuery(TestCase):
         self.client.post(
             url,
             data={"email": "foo@bar.com"},
-            **{"HTTP_X_REQUESTED_WITH": "XMLHttpRequest"}
         )
         self.assertEqual(mocked_execute.delay.call_count, 1)
 
-    def test_email_403(self):
-        query = SimpleQueryFactory()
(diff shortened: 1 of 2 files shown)


jarcoal:
Bumping this

chrisclark:
yeah totally unclear why this check is here! will merge!

jarcoal:
Thanks @chrisclark .  Looks like the check that failed is unrelated to this PR.

